PR #115, opened 2013-05-02
Wildcard matching option for query string parameters

paullryan:
This is a wildcard matching change so that /path/:to/:foo?\* is an acceptable url from the UrlMatcherFactory. If given this will match all query string parameters and pass that down to the state. This will also by default cause a reload. You can set the option in your state for reloadWild: false to disable the auto reload based on changes here. 

Please feel free to comment about what else you might like to see here as this meets my needs but I'm not sure it meets all needs that I've seen moving back and forth on this issues list. I'd be happy to update this for those needs or give someone else collaborator access to this branch if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ node_modules
 
 # webstorm files
 .idea
+*.iml
--- a/src/state.js
+++ b/src/state.js
@@ -178,7 +178,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
       // But clear 'transition', as we still want to cancel any other pending transitions.
       // TODO: We may not want to bump 'transition' if we're called from a location change that we've initiated ourselves,
       // because we might accidentally abort a legitimate transition initiated from code?
-      if (to === from && locals === from.locals && to.params.indexOf('*') < 0) {
+      if (to === from && locals === from.locals) {
         $state.transition = null;
         return $q.when($state.current);
       }
@@ -287,12 +287,14 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
       // we also need $stateParams to be available for any $injector calls we make during the
       // dependency resolution process.
       var $stateParams;
-      if (paramsAreFiltered && params['*'] === undefined) $stateParams = params;
+      if (paramsAreFiltered) $stateParams = params;
       else {
         $stateParams = {};
         forEach(state.params, function (name) {
           if(name === '*'){
-              // TODO ??? needed for transition but what...
+              for(var key in params){
+                  $stateParams[key] = params[key];
+              }
           } else {
             $stateParams[name] = params[name];
           }
@@ -355,6 +357,9 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
     }
 
     function equalForKeys(a, b, keys) {
+      if(keys.indexOf('*') > 0 && a != b){
+          return false;
+      }
       for (var i=0; i<keys.length; i++) {
         var k = keys[i];
         if (a[k] != b[k]) return false; // Not '===', values aren't necessarily normalized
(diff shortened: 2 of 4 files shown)


frogomatic:
Arbitrary query parameters is a big deal in our app which utilizes backend search faceting. It's not feasible to register the parameters on the frontend since they can change at runtime. Is there any movement on this or a workaround?

cheahkhing:
Hi, i also needed this changes for my requirement of having dynamic query parameters being passed in the url. Is this going to be merged into master? When will i get this in the release?

Schlesiger:
Please implement this. It would be very useful to my project. For now my only option is hardcoding field names which are defined in the backend.

nathanredblur:
When will be this ready?

nateabele:
@nathanredblur No one is working on it right now.

frogomatic:
I wound up patching our local version. With the patch you can use
QUERYPARAMS in a state definition e.g.

$stateProvider.state('mystate', { url: '/app/something/:query?QUERYPARAMS',
... });

This will set the QUERYPARAMS parameter to the JSON encoded query
parameters. Any other defined parameters are also set.

You can also do a

$state.go('mystate', {query: 'something', QUERYPARAMS:
JSON.stringify(myQueryParamsHash)});

to trigger a state change.

It's a kludge, but it works for our needs.

Patch is attached. I'm not sure how well github handles attachments, so if
you don't get it, let me know. The patch applies cleanly to 0.2.14 and
0.2.15.

robwalkerco:
@frogomatic I donk think the patch is available from the attachment. Please could you make it available another way? Do you have a fork?

frogomatic:
Yes, I have a fork: https://github.com/frogomatic/ui-router

The 'master' branch contains the QUERYPARAMS workaround. I applied the
patch to angular-ui-router.js, and manually to the source files, but not to
the minified js file.

There's also a 1119-workaround branch which contains a partial workaround
for the double-encoding of slashes in state parameters (issues 1119 / 1645)

kindy-aa:
how about add `$stateSearch` & `$stateHash` instead of change `$stateParams`

kindy:
For my case, I can get param list in `state's resolve`, then I can update current state's `searchParams` dynamic. then trigger `$urlRoute.sync()` again. and now my state will get correct `$stateParams` .

As I see, this `state.updateSearchParam` need to update `state.url`, `state.OwnParams` & `state.Params`.
currently `UrlMatcher` dose not support `update`. and the `contact` not work for this case.

paullryan:
I know I've let this languish as the core team seemed a little against my initial attempt. If they have a suggestion for pushing this forward such as pulling in the changes from @kindy I'd be happy to help move this issue along. Any thoughts from the core team?

christopherthielen:
I think we're OK with the wildcard query parameter as a general idea.   However, internal parameter handling has had a major overhaul in 1.0, and will continue to change until 1.0 final is reached.  This PR will not be compatible with that branch, and parameters in that branch are a bit of a moving target at this time.

Before we'll accept a wildcard PR, we need to address the original caveat raised by ksperling back in 2013.  In order to determine which states in the toPath (if any) we should "keep" or we should "reload", we must associate the raw parameters with the states they belong to, and then compare the values.

We need to ensure that the wildcard query param does not interfere with mapping params to path parameters, and does not interfere with mapping query parameters to parent states, as well as to potential substates.

This may be challenging, because the parameters which are associated with a state are statically computed as discrete `Param` objects.

Here are some relevant lines:
https://github.com/angular-ui/ui-router/blob/f8e5f671eda0114becf5e618ceb4350ac01ab926/src/path/node.ts#L23-L26

https://github.com/angular-ui/ui-router/blob/f8e5f671eda0114becf5e618ceb4350ac01ab926/src/path/node.ts#L40-L41

ergo:
Any update on this? I have a search system that accepts arbitrary "tag names" as params so I cant specify all params upfront in the urls.

peterjhart:
**BTW** if you want to access query params you can still inject $location and grab them through $location.search().

ergo:
@peterjhart, Yes, but the problem is if you don't specify param names in url - the state is not being reloaded. And if memory serves me right, also ui-sref will ignore additional params when generating urls.

peterjhart:
@ergo ohisee. D'oh.

PR #446, opened 2013-09-18
Add optional parameter matching to $state.is and $state.includes

### LINKED ISSUE #413: $state.is and $state.includes should support second arg for id
``` html
<li ng-repeat="chapter as chapters" ng-class="{ active: $state.includes('book.chapters', chapter.id) }"><a ui-sref="book.chapters({ chapterId: chapter.id})">{{chapter.name}}</a></li>
```

Just a little sugar to make these types of state matching a little nicer.

roryf:
Fixes #413.

Allows passing a second parameter to `$state.is` and `$state.includes` with parameters for the state. `is` will do a strict `angular.equals` on the passed parameters with the current `$stateParams`, while `includes` will only attempt to match the parameters given.

Examples:

```
{ active: $state.is('site.section', {siteId: site.id, sectionId: section.id)} }
```

```
{ active: $state.includes('site.section', {sectionId: section.id)} }
```

I also changed the method contract so it won't return `undefined` when a state doesn't exist. This didn't make sense to me that a function could return a boolean value, or `undefined`, especially when the docs state it will return a boolean.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -377,14 +377,37 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory,           $
       return transition;
     };
 
-    $state.is = function is(stateOrName) {
+    $state.is = function is(stateOrName, params) {
       var state = findState(stateOrName);
-      return (isDefined(state)) ? $state.$current === state : undefined;
+
+      if (!isDefined(state)) {
+        return undefined;
+      }
+
+      if ($state.$current !== state) {
+        return false;
+      }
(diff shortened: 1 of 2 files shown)


vdevappa:
Can someone please add support for the $ne operator for the JSON in Angular 1.x please?

`{ active: $state.includes('site.section', {sectionId : { $ne: section.id } )} }`

I have a situation where it needs to be active when it does not match a particular value.

PR #454, opened 2013-09-20
Parameter Typing -- URL Only

toddhgardner:
I'm a bit concerned I'm not going down the right path here, so I'd love some feedback on this before I go farther. With this code, you can add types which will be encoded/decoded as they pass back and forth to the URL. Everything is handled with the URL property right now, no params structure.


phazei:
Can someone point to the docs for this?  I can't find anything for where parameter typing is :(

PR #528, opened 2013-10-25
Prepend the baseHref to $state.href() generated urls if it exists.

tgecho:
I have an angular app that is served from a subdirectory. I have a `<base href="/app/">` tag, however the urls generated by the `ui-sref` directive and `$state.href()` do not reflect this. The ui-sref links themselves will still work if left clicked, however middle clicking to open in a new tab does not work since the url is missing the contents of the base href.

I think this pretty much covers things the various scenarios...


samy100:
Is there any option I can turn automatic prefixing of the base URL off? or make it configurable ui-sref-option like "prefixBase: false"? Thanks!

Issue #510, opened 2013-10-17
Using $location.hash()


acroca:
Hi everyone,

First of all, thanks for this great project!

Is there any way of sending `$location.hash()` to a state?

I've checked the code and the default `options.location` when going to a new state is `true`, so it rewrites the location getting rid of the hash. Also I tried to create a state with a url like `/:id#:section` but didn't work.

In case this is still not supported what do you think it's the best approach to define a state with hashes in the url?

Thanks

eddiemonge:
https://gist.github.com/eddiemonge/f6a58169c2846731a1eb

mssalnikov:
As per note, I'm commenting to say I'm still affected by this issue, since I need to match urls like `/{param1:[0-9]+}/{param2:#?[0-9]+}

richfergus:
@mssalnikov,
I got the $state to work in this:

        $scope.flightHours = function() {
            $state.go('app.ops.main',{'#':'red'});
         };

With this anchor tag on the ops.main template:
       
        <div class="row" id="red">

mssalnikov:
@richfergus Hey! 

Yes, I'm aware of that solution, yet it only works from withing angular, when you change the state manually. My case is this: users have to be able to access url https://domain.com/3456/#1234, and $state does not match this to a `/{param1:[0-9]+}/{param2:#?[0-9]+}`. My workaround is therefore ugly AF:

`
.state('catch', {
            url: '/{param1:[0-9]+}/',
            onEnter: /* @ngInject */ ($state) => {

                if (location.hash) {
                    $state.go('state', {param1: $state.params.param1, param2: location.hash})
                }
            }
        })
`

PR #612, opened 2013-11-27
Preserve url fragment on page reload

beyang:
This fixes a bug where the URL hash disappears when a page is first loaded.

For instance, if you had a link `<a href="#someanchor">link</a>` on your page and clicked on it, the link would work correctly (it would scroll to `#someanchor`), but if you copied and pasted the URL into a new tab and navigated to the page afresh, the hash would disappear and no scrolling would occur.

See also: https://github.com/angular-ui/ui-router/issues/510

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -181,7 +181,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory,           $
     if (!state['abstract'] && state.url) {
       $urlRouterProvider.when(state.url, ['$match', '$stateParams', function ($match, $stateParams) {
         if ($state.$current.navigable != state || !equalForKeys($match, $stateParams)) {
-          $state.transitionTo(state, $match, false);
+          $state.transitionTo(state, $match, { location: false });
         }
       }]);
     }
--- a/test/stateSpec.js
+++ b/test/stateSpec.js
@@ -689,6 +689,17 @@ describe('state', function () {
       expect($state.current.name).toBe('about.person');
     }));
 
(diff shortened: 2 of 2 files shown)


cyberwombat:
Are there docs on how to enable this? I have the latest UI Router and named anchors do not work for me when loaded from address bar.

nateabele:
@cyberwombat I don't think you're asking for the same thing as this PR addresses. What do you mean by named anchors?

cyberwombat:
@nateabele  - seems like the original entry to this issue describes it. Basically my nameed anchors, i.e. `<a name="about"></a>` work fine if I have a link on the page that looks like `<a href="#about">About</a>`. This causes the page to shift to location as well as update the URL. If I press reload at that point (which would be the same as a bookmark/paste as original posted mentions) then it does not find the page anymore and 404s. I have tried adding some basic regex to the router but my results were not successful. I would end up with things like `http://example.com/about` instead of the desired `http://example.com#about` or even `http://example.com`. Ideally I would like the hash anchor to be preserved but would be ok with it stripped away as long as things dont break. It seems OP question was about that but i am not sure the PR was - may be about the non html5 fragments.

nateabele:
> If I press reload at that point (which would be the same as a bookmark/paste as original posted mentions) then it does not find the page anymore and 404s.

This sounds like a server configuration issue, which would be unrelated to UI Router. Please see https://github.com/angular-ui/ui-router/wiki/Frequently-Asked-Questions#how-to-configure-your-server-to-work-with-html5mode

PR #678, opened 2013-12-13
add view name to $viewContentLoaded arguments

tamtakoe:
$viewContentLoaded is cool event, but it will be better and useful if we can get view name within handler


vladgurovich:
It looks like the reason this PR is ignored is because it has no test coverage

nateabele:
No test coverage, needs to be rebased against master, needs to follow contributor guidelines.

eddiemonge:
@tamtakoe ^^

nateabele:
Fixed in #2003.

PR #927, opened 2014-03-04
feat(uiSrefActive): Also activate for child states.

### LINKED ISSUE #818: ui-sref-active should be able to match on nested child states.
Feature Request:

Given a state 'foo' and child state 'foo.bar' ui-sref-active should be able to maintain it's 'active' status on the parent state, while the child state is in view. 

plnkr demonstrating usefulness: 
http://plnkr.co/edit/UaBV3DPwR6qt3KLTjQ0r?p=preview

Use navigation menu to hit manage flights
Note two flights F-0000, F-0001 neither of which are listed as "active"
Click on one of them.
Border is drawn around selected tab.
Now click on the "search w/ Criteria" button inside the active state.
The "active" state is lost on the selected tab, even though the displayed state is a child of the tab. 

Discussed with Caitp in irc.

timkindberg:
feat(uiSrefActive): Also activate for child states.

To limit activation to target state use new ui-sref-active-equals directive'

Breaking Change: Since ui-sref-active now activates even when child states are active you may need to swap out your ui-sref-active with ui-sref-active-equals, thought typically we think devs want the auto inheritance.

Fixes #818

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -80,7 +80,7 @@ function $StateRefDirective($state, $timeout) {
 
   return {
     restrict: 'A',
-    require: '?^uiSrefActive',
+    require: ['?^uiSrefActive', '?^uiSrefActiveEq'],
     link: function(scope, element, attrs, uiSrefActive) {
       var ref = parseStateRef(attrs.uiSref);
       var params = null, url = null, base = stateContext(element) || $state.$current;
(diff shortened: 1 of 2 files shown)


fxck:
Is this case gonna cover it?

I have a link in menu that points to `app.something.list()`, but I want it to be active even when I'm on `app.something.detail()`..

maciej-gurban:
+1 for what @fxck suggests.

wcandillon:
@fxck @maciej-gurban I've made a proposal for this at https://github.com/angular-ui/ui-router/pull/1254 What do you think?

fxck:
I'd say this should be opt out rather than opt in, and if opt in, then not by explicitly naming the states. What you proposed is basically the same as using ngclass..

uguryilmaz:
+1

PR #1021, opened 2014-04-11
feat($state): event when preventing state transition to self

### LINKED ISSUE #1016: Event when transition to same state is cancelled
There is currently a lack of an event to mark the end of the state change when transitioning to the same state. To be exact, the `$stateChangeStart` event is emitted, but no event signifying the end of the transition (neither `$stateChangeSuccess` nor `$stateChangeError`). This makes it impossible to cover all possible transition results via event handling.

To be precise, [this code](https://github.com/angular-ui/ui-router/blob/master/src/state.js#L811) cancels the transition: 

```
      if (shouldTriggerReload(to, from, locals, options) ) {
        if (to.self.reloadOnSearch !== false) $urlRouter.update();
        $state.transition = null;
        return $q.when($state.current);
      }
```

I'd like to suggest that an event would be sent, something along the lines of `$stateChangeCancelled` or something of the sort.

nikaspran:
Fire a `$stateChangeToSamePrevented` event when transitioning to self. 

This adds an end event broadcast for state change scenario A->B->A (currently, the `$stateChangeStart` event would be broadcast but neither `$stateChangeSuccess` nor `$stateChangeError` would signal the completion - meaning you can't cover the entire state transition via event handling).

Closes #1016

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -811,6 +811,23 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
       if (shouldTriggerReload(to, from, locals, options) ) {
         if (to.self.reloadOnSearch !== false) $urlRouter.update();
         $state.transition = null;
+        if (options.notify) {
+          /**
+           * @ngdoc event
+           * @name ui.router.state.$state#$stateChangeToSamePrevented
+           * @eventOf ui.router.state.$state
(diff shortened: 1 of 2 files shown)


marshall007:
@nikaspran I think this will finally be solved once #1090 lands!

PR #1023, opened 2014-04-11
feat($state): Make variable 'states' public in $stateProvider

maxaon:
Allow access to variable `states` in `$stateProvider` for extending $state using angular's [decorator](http://docs.angularjs.org/api/auto/object/$provide#decorator)
This need to extend `resolve` property of state.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -23,6 +23,7 @@ $StateProvider.$inject = ['$urlRouterProvider', '$urlMatcherFactoryProvider'];
 function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
 
   var root, states = {}, $state, queue = {}, abstractKey = 'abstract';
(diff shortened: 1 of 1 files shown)


wesleycho:
Closing, as per comments & not enough justification as to why `$provide.decorator` is not sufficient for wanting to expose internals.

In addition, this lacked tests.

christopherthielen:
> I've putted example previously: http://jsbin.com/befod/15/edit
> I want create method goResolve which adds dynamic injectable parameters to controller with promise resolving.

feature-1.0 branch supports this in the form of transition hooks, by the way.   Something like this:

```
function makeDynamicResolves(MyService) { 
  return {
    newResolve: MyService.fetchSomethingAsync()
  }
}

$transitions.onStart({}, makeDynamicResolves)
`

PR #1037, opened 2014-04-22
Adding 'reloadOnPath' and 'reloadOnQuery' configuration to the state.

cbayram:
The reloading and recreation of the controllers and views when modifying search parameters prohibits any smooth UI as the DOM as well as scopes are discarded. The current 'reloadOnSearch' stops the transition, however it fails to update the $location.search(), more importantly, it fails to reflect the changes to the $stateParams which applications rely on for reactive code ($watch). 

The configuration of 'reloadOnPath' and/or 'reloadOnQuery' being set to 'false' will update the url, update the $stateParams and suppress the reloading of the controllers and views. One can specify the same configurations when using $state.transitionTo() as well as other wrappers around it ($state.go()).

### DIFF
--- a/test/stateSpec.js
+++ b/test/stateSpec.js
@@ -27,6 +27,9 @@ describe('state', function () {
       HH = { parent: H },
       HHH = {parent: HH, data: {propA: 'overriddenA', propC: 'propC'} },
       RS = { url: '^/search?term', reloadOnSearch: false },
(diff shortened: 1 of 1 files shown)


rolos79:
@nateabele I'm very keen to use the functionality you added in the `dynamic-params` branch back in April 2014, but am apprehensive has it hasn't been added to a release yet. Are you aware of any reasons why it hasn't been included yet?

PR #1078, opened 2014-05-15
fix(uiSref): fix to uiSref when using dynamic state names

### LINKED ISSUE #395: Dynamic ui-sref in directive
I've written a breadcrumb directive which takes an array of crumbs, each with a name and sref property. I'm trying to bind the `ui-sref` directive (`$StateRefDirective`) to use the evaluated attribute, but it's not working. When the `ui-sref` directive's `link` function is called, the `attrs.uiSref` value is `undefined`, but immediately after the running of the function becomes available as the correct value.

```coffee
angular.module('myApp').directive 'breadcrumb', ->
  scope: { crumbs: '=' }
  replace: true
  template: '
    <nav class="breadcrumbs">
      <a ui-sref="{{crumb.sref}}" ng-repeat="crumb in crumbs">{{crumb.name}}</a>
    </nav>
  '
```

Inside the `$StateRefDirective` `link` method, logging `attrs` gives

```js
Attributes {… uiSref: undefined …}
```

Running the same log after a 100ms timeout gives: 
```js
Attributes {… uiSref: "apps.show({appId: crumb.app.id})" …}
```

Am I doing something wrong, or are dynamic `ui-sref` attributes not supported?

Note: Running `v0.2.0`

lee-tunnicliffe:
Fix to ensure href updates when the state name or the params change as
opposed to just the params.

Closes #395

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -115,6 +115,11 @@ function $StateRefDirective($state, $timeout) {
         element[0][attr] = newHref;
       };
 
+      attrs.$observe('uiSref', function(newVal) {
+        ref = parseStateRef(newVal);
+        update();
+      });
+
       if (ref.paramExpr) {
         scope.$watch(ref.paramExpr, function(newVal, oldVal) {
           if (newVal !== params) update(newVal);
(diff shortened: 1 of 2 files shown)


nateabele:
Closing this as we decided to roll with an alternate directive.

PR #1090, opened 2014-05-21
adds statechangecancel event

elmerbulthuis:
After preventDefault is fired in stateChangeStart, a stageChangeCancel event will be launched.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "0.10.17"
+  - "0.10"
 
 before_script:
   - export DISPLAY=:99.0
--- a/src/state.js
+++ b/src/state.js
@@ -847,6 +847,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
          * </pre>
          */
         if ($rootScope.$broadcast('$stateChangeStart', to.self, toParams, from.self, fromParams).defaultPrevented) {
+          $rootScope.$broadcast('$stateChangeCancel', to.self, toParams, from.self, fromParams);
           $urlRouter.update();
           return TransitionPrevented;
         }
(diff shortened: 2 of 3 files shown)


rpocklin:
Any reason this hasn't landed yet?  I need this and am working around it for the time being.

christopherthielen:
While the state events are probably going to be deprecated in favor of transition callbacks in 1.0, I think this is a reasonable thing to include in 0.2.x, for those who can't (or won't want to) update to 1.0.  @nateabele  what say you?

nateabele:
I was always kind of against this because you're basically notifying your own code of what it's doing. I'm willing to be persuaded on this point though.

rpocklin:
If you visit http://rpocklin.github.io/angular-ui-view-spinner/example/index.html and click some links you can see it's a directive that shows a loading spinner while the `resolves` are processed.  There are 2 ways (AFAIK) that route state transitions in `ui-router` can fail:

* A resolve for the target state throws an exception (or returns a falsy value)
* In a `$stateChangeStart` event handler, you can do `event.preventDefault();`

In the first case, a `$stateChangeError` is triggered.  In the second, though, there is no event triggered such as `$stateChangeCancel` which means the loading spinner has no hook to stop.

I can manually trigger a `$stateChangeError` event for the second case, but IMO this is a prime candidate for a different event, especially since it's the user or system deciding to cancel the state change, and not because a dependent resolve has failed.

@nateabele if you are writing both sets of code within the same codebase, you are right you have control over the interactions.  But what if you are combining two third party directives which don't directly talk to each other?  Eg. the above one and one which pops up a window to prevent unsaved changes in a state being lost? (Eg. https://github.com/facultymatt/angular-unsavedChanges)

nateabele:
That's fair. Squash & fix the commit message per the contributor guidelines and we'll merge it.

rpocklin:
Done - see #1844 as original PR wasn't mine.

nateabele:
Implemented in #1844.

Issue #1095, opened 2014-05-22
Ability to remove states from $stateProvider


JensEger:
We are perform lazy state binding. This brings the advantage, that a authenticated (or even a unauthenticated) User can "see" only his states / routes, which is also a security reason.

When a user logs out, i am not able to clear the lazy registered routes. I would like this to be implemented. This could be a really interessting feture, because i saw this lazy registration in other SPA's .

Suggested usage: 
```
$stateProviderService.removeState('statename')
```

christopherthielen:
Closed by 44579ec

PR #1114, opened 2014-06-02
add removeState() to $stateProvider [feature] #1095

ilyagelman:
https://github.com/angular-ui/ui-router/issues/1095


eddiemonge:
Can you give use cases for this?

paulflo150:
I would like to see this implemented as well, as it would be very useful for states that are being added via a provider upon a user's login.

@eddiemonge, here is the way my app works: 
1. I have a couple of default states a user can access without having to login which I hard coded in the config phase.
2. Upon a successful login, and based on the user's roles I retrieve the states this particular user has access to, and add them via a provider.
3. So far all is well, but when the user logs out, and if another user logs in without performing a hard refresh, one cannot clear out/ override any of the previous states.

Is there perhaps a way to remove ALL the routes that I may have missed in the docs?

Assuming there isn't, what are the chances we will be seeing a removeState anytime soon?

Thanks!

eddiemonge:
Why not use the state's data option and specify what role is needed to access the state. in the stateChangeStart event, check if the user has that role and if not then you take them to an error page? Then you don't need to muck about with adding/removing states.


Closing this because it needs a lot of work. 
 * The release files were edited instead of src
 * Its missing tests
 * The commit message doesn't follow the contributing guidelines

paulflo150:
One could do that, but having to maintain roles in two places, and hard code them in the UI layer seems like a poor design.

RichardFans:
we need this feature ! eddiemonge, please give us remove removeState()

vinothbabu:
@paulflo150 : This would be rarely used, i am just thinking the scenarios... if a user removes a state via removeState(), it would be there only until the application lifecycle, once its app is reloaded, your state is still accessible. There should be somebackend mechanism needed for persistance. 

Why not use resolve

`.state('someState', {
                  url: url,
                  templateUrl: template,
                  controller: 'Ctrl'
                  resolve: {
                      permission: function (Factory, $window) {
                          return Factory.checkPermission($window.sessionStorage.sendSomething);
                      }
                  }
 })`

Inside your checkPermission method, you can see whether he can access or not... based on that you can redirect him to other state. I am not against adding this method, just thought would throw in some my thoughts.

Issue #701, opened 2013-12-19
RFC: support URL fragments in ui-sref


beyang:
Proposal is to support \# URL fragments/hashes in ui-sref.  So you could do:
`<a ui-sref="mystate(myparams)#myhash>...</a>`

It appears that `$state.go(href)` already supports having a hash in href, so all this would require is making a change to the ui-sref directive (make it parse the hash and then pass this to $state.go).

cryode:
For anyone else struggling with this particular issue, here's how I went around it:

```js
angular.module('app').config(['$provide', function ($provide) {

  /**
   * Extend the UI Router $stateProvider, adding the ability to specify an
   * anchor hash as a parameter in the ui-sref directive.
   */
  $provide.decorator('$state', ['$delegate', function ($stateProvider) {

    // Save the orignal function for generating a state's URL.
    var $stateHref = $stateProvider.href;

    // Create our extended function.
    $stateProvider.href = function href(stateOrName, params, options) {
      var hash = '';
      params = params || {};
      var hashParam = params['#'];

      // Check if the anchor parameter was specified.
      if (typeof hashParam !== 'undefined') {
        // Ensure hash parameter is a string and not empty.
        if ((typeof hashParam === 'string' || hashParam instanceof String) && hashParam.length) {
          hash = '#' + hashParam;
        }

        delete params['#'];
      }

      // Return the original parsed URL with the hash appended.
      return $stateHref(stateOrName, params, options) + hash;
    };

    return $stateProvider;
  }]);

}]);
```

[`$state.href`](https://github.com/angular-ui/ui-router/blob/0.2.13/src/state.js#L1273-L1294) is where the state identifier / params / options are parsed into a path (as best as I can tell, anyway). The code is pretty easy to read, but the workflow is pretty much:

1. Original function is saved, so it can be used again (not sure if there's a more OOP way to do this -- thoughts?)
2. Checks if there is a `#` parameter supplied. If so, save its value, then remove it from the params object.
3. Generate the path using the original function.
4. Return the generated path with the anchor appended.

This is used by simply specifying a `#` parameter as part of your `ui-sref` directive:

```html
<a ui-sref="my.state({ id: Model.id, '#': IDToScrollTo })" class="btn">
```

mismith:
[...]

PR #1187, opened 2014-07-04
support setting url #fragments in state params

mismith:
As per @sadlerw's request in #701

NB: obviously HTML5 mode needs to be enabled for this to have any effect
Supports both formats:
- `ui-sref="mystate(myparams)#{{ myhashexpression }}"`
- `ui-sref="mystate({myparam1: myvalue1, myparam2: myvalue2, '#': myhashexpression})"`

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -815,7 +815,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
       }
 
       // Filter parameters before we pass them to event handlers etc.
-      toParams = filterByKeys(objectKeys(to.params), toParams || {});
+      toParams = filterByKeys(objectKeys(to.params).concat('#'), toParams || {});
 
       // Broadcast start event and cancel the transition if requested
       if (options.notify) {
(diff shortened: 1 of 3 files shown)


OnkelTem:
Murray, would you provide some example on how this is supposed to work?
Specifically, should location be rewritten after state transition to include #<fragment>?  From what I test this is not happening.

mismith:
@OnkelTem: "should location be rewritten after state transition to include #<fragment>". Yes.
Perhaps since I created this PR the `master` branch has changed in such a way that this no longer works—or maybe browsers have evolved themselves—but back then it did indeed add the #fragment to the URL after/on clicking the link.

mismith:
Yes, it looks like the following line in the latest `master` branch is what's stripping out the '#' param:

    toParams = toState.params.$$values(toParams);

The `$$values` prototype method does some internal filtering magic, so by the time we get to building a URL in the parent `$state.transitionTo` method, we no longer have `toParams['#']` set.

Issue #1100, opened 2014-05-27
Adding custom previous property to the $state object


thak123:
Ability to remember the previous state from which the current state was reached ...

Is there a way to add custom previous property to the $state object in the controller when i am changing the the State...

For ex i am on state A going to state B and i can go to state C from state B also from D i can go to State C ,

i want to remember from which state i jumped to state C ..i.e either B or D 

So may be we can have  a construct which takes in the custom properties and which can be used later when going back..

So i will just say $state.goBack()

eddiemonge:
https://gist.github.com/eddiemonge/f6a58169c2846731a1eb

PR #1218, opened 2014-07-23
feat($state): expose previous state/params via $state

### LINKED ISSUE #92: Access previous state
To my knowledge this is not supported. It would be very helpful to be able to know the previous state, even if it is just the state's name. This information is useful for when you need to do logic based on which state you reached the current state from.

homerjam:
Please please please, fixes #92 and #1100


amitagrawal11:
I have created my own service to handle this. Please check it out and let me know if any concern comes

https://github.com/amitagrawal11/UI-Router-History

Issue #1220, opened 2014-07-24
Custom urlMatcherFactoryProvider types not returning null when not matching


tiwatson:
Based on the code example in:
https://github.com/angular-ui/ui-router/blob/master/src/urlMatcherFactory.js#L602

Should this test pass or should it not? 
It fails with: Expected { foo : undefined } to be null.

```
it("should match custom types and return null when not a match", function () {
  var list = ['a'];
  var type = {
    decode: function(item, v) {
      return list[list.indexOf(item)];
    }
  };
  $umf.type("myType", type);
  expect($umf.compile("/{foo:myType}").exec("/a")).toEqual({ foo: 'a'});
  expect($umf.compile("/{foo:myType}").exec("/b")).toBeNull();
});
```

This test based on the special int type does pass

```
it("should match number in int and not a letter", function () {
  expect($umf.compile("/{foo:int}").exec("/1")).toEqual({ foo: 1});
  expect($umf.compile("/{foo:int}").exec("/b")).toBeNull();
});
```

UrlMatcher.prototype.exec = function (path, searchParams) {
This functions description states:

 * Tests the specified path against this matcher, and returns an object containing the captured
 * parameter values, or null if the path does not match. The returned object contains the values
 * of any search parameters that are mentioned in the pattern, but their value may be null if
 * they are not present in `searchParams`. This means that search parameters are always treated
 * as optional.
 
Since null is returned for any non matched Regex Parameters and null is returned for the special int/bool/date types, shouldn't null be returned if a custom type is not matched?

eddiemonge:
https://gist.github.com/eddiemonge/f6a58169c2846731a1eb

PR #1221, opened 2014-07-24
Add nullOnNoMatch config option for custom  types

tiwatson:
Related to the issue i created here: 
https://github.com/angular-ui/ui-router/issues/1220

Not sure this is the best way to go about it. Open to thoughts to make it better.


nateabele:
Yeah, not a fan of adding configuration flags for every imaginable use case. If you're already defining a custom type, you can easily handle this on your own.

PR #1254, opened 2014-08-06
Add ui-sref-active-includes attribute

### LINKED ISSUE #1252: Bug in "How to: Set up a default/index child state"
https://github.com/angular-ui/ui-router/wiki/Frequently-Asked-Questions#how-to-set-up-a-defaultindex-child-state

This solution doesn't appear to work with `ui-sref-active`. Going the default state will cause `ui-sref-active` to not highlight siblings of the default state. Going to the abstract state will cause an error.

wcandillon:
fixes #1252

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -200,6 +200,26 @@ function $StateRefDirective($state, $timeout) {
  *   </li>
  * </ul>
  * </pre>
+ * 
+ * @example
+ * Given the following template where `app.users` is an abstract state and `app.users.default` its default child state:
+ * <pre>
+ * <ul>
+ *   <li ui-sref-active="active" ui-sref-active-includes="app.users" class="item">
+ *     <a href ui-sref="app.users.default">Users</a>
+ *   </li>
+ * </ul>
+ * </pre>
+ * 
+ * When the app state is "app.users" (or any children states),
+ * the resulting HTML will appear as (note the 'active' class):
+ * <pre>
+ * <ul>
+ *   <li ui-sref-active="active" ui-sref-active-includes="app.users" class="item active">
+ *     <a ui-sref="app.users.default" href="/users/default">Users</a>
+ *   </li>
+ * </ul>
+ * </pre>
  */
 
 /**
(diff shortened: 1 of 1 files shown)


dougmoscrop:
@christopherthielen but that's precisely what lots of people go and do. you have a top navbar, with a bunch of links to states (foo.index, bar.index, baz.index) and then within those there's a lot of child states of foo, bar and baz (foo.details, bar.moreinfo, baz.blahblah) but they do not descend from the index state. and we want the links to the top states to show active because you're in that 'section' of the app.

how about `ui-sref-active-parent` as an alternative way of expressing this? it doesn't require the explicit route reference, but it will activate if the parent state of the ui-sref is active.

so as an example, given a ui-sref to `foo.bar.baz.index`,

the following states would *not* apply active:

- `foo.*`
- `foo.bar.*`

but any state that contains `foo.bar.baz` would, e.g. `foo.bar.baz.index`, `foo.bar.baz.detail`, `foo.bar.baz.detail.more.lots.really.toomuch`

nateabele:
Also: https://radify.github.io/PR.js/#/angular-ui/ui-router/pulls/1254

eddiemonge:
@wcandillon Can you clean up the PR please?

What about `ui-sref-in-active`

christopherthielen:
I think this bug would be irrelevant if we had proper "redirect to default substate", yes? 

```
<a ui-sref="app" ui-sref-active="active">Go to App Default Substate</a>
```

```
$stateProvider.state('app', { redirectTo: 'app.index' })
.state('app.index', { url: '/' })
.state('app.details', { url: '/:id'})
```

```
app.run(['$rootScope', '$state', function($rootScope, $state){
      $rootScope.$on('$stateChangeStart', function(event, toState) {
        var redirect = toState.redirectTo;
        if (redirect) {
          event.preventDefault();
          if(angular.isFunction(redirect))
              redirect.call($state);
          else
              $state.go(redirect);
        }
      });
    }]);
```

cc: @wcandillon @dougmoscrop @eddiemonge

dougmoscrop:
Yes, assuming that 'app' is still a valid abstract state etc

gercheq:
I totally agree with @dougmoscrop I always have an abstract state for each section of my app as I'm doing some work under the hood for each section. Unfortunately I can't link to an abstract state and linking to one of the substates does not trigger ```.active``` on the navigation item.

christopherthielen:
For people searching and finding this issue:

1) We now have [first class support for redirects](https://ui-router.github.io/docs/latest/interfaces/state.statedeclaration.html#redirectto) which makes it easy to implement "default substates".

2) Although it doesn't seem to be generated in the docs properly, we have support for  [state name globs in ui-sref-active](https://github.com/angular-ui/ui-router/blob/6ac1c61/src/stateDirectives.js#L267-L283), e.g., `ui-sref-active="{ 'active': 'admin.**' }"`

PR #1521, opened 2014-11-07
Allow multiple states to be registered at once

dotnetwise:
e.g 
```js
$stateProvider.state([{
  name: 'a',
  url: 'a',
}, {
  name: 'b',
  url: 'b',
}]);
```
This is the way all other angular directives / config etc. coding style work


ilanbiala:
@dotnetwise doesn't seem very useful because the fluent syntax is pretty easy to read already.

dotnetwise:
It is very useful because it's consistent with all other Angular stuff

PR #1585, opened 2014-11-26
fix($urlRouter): resolve clashing of routes

just-boris:
Sometimes url can be matched by several different rules. Now it will be handled by the first rule in order of appearance. So it causes some issues. It would be better to match states from more specified to common.

Angular-route already has same issue https://github.com/angular/angular.js/issues/8266

So, I propose include this fixes here, because is ui-router more powerful and flexible library.

### DIFF
--- a/src/urlRouter.js
+++ b/src/urlRouter.js
@@ -303,6 +303,12 @@ function $UrlRouterProvider(   $locationProvider,   $urlMatcherFactory) {
       return listener;
     }
 
+    rules.sort(function(ruleA, ruleB) {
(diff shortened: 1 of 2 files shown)


wesleycho:
What are your thoughts on this @nateabele ? This sounds sane, but there is still some ambiguity with states with the same level of specificity.

nateabele:
@wesleycho I don't actually remember what I was gonna do here, but this solution seems fine for now.

nateabele:
@wesleycho Btw, any PRs that get merged to master from here out, be sure to flag them on the 1.0-preview milestone so we can review. /cc @christopherthielen

PR #1613, opened 2014-12-06
feat(uiSrefActive): Added support for multiple nested uiSref directives

dtritus:
Now `uiSrefActive` will add active class if match any state of nested `uiSref` directives.


dtritus:
I've rebased the PR against the latest changes

nateabele:
Patch looks great, well done. Thanks for your work on this.

Issue #1657, opened 2014-12-24
$element not injected as a local


Nate-Wilkins:
I'm really confused as to why ui-router controllers aren't allowed access to `$element`? Is this by design?

[Plunker](http://plnkr.co/edit/g28HkEeN4xAM6YX0dW5J?p=preview)
[Related Code](https://github.com/angular-ui/ui-router/blob/2e718534e889c613958a71e23e5637059676c177/src/viewDirective.js#L277)

In the end I need to end up accessing a directive controller from my controller like so:

``` js
app.controller('MyController', function ($element) {
    var cool = $element.inheritedData('$coolController');
});
```

PR #1659, opened 2014-12-25
fix(uiView): add `$element` to locals for view controller

Nate-Wilkins:
Closes issue #1657


Nate-Wilkins:
@nateabele When is the next minor release? (Since this isn't in the latest release)

Issue #1667, opened 2015-01-03
ui-sref inside inlcuded svg stopped to work since 0.2.11


martinoss:
I included an SVG image with some links (ui-sref) in my project using ng-include.
After I upgraged from 0.2.10 to 0.2.13, the links are no more getting invoked. It turned out that it's broken since 0.2.11.

Here is an example:
http://plnkr.co/edit/uod9jpkySgYsXya66v51?p=preview
If you change the included ui-router script to 0.2.11 or 0.2.13, clicking the svg has no effect.

The issue should be easy to fix:
In the directie $StateRefDirective there is a check if the element is an anchor element:
```isAnchor = element.prop("tagName") === "A";```
In case of the SVG, the tag name was lowercase ("a"). 
```isAnchor = element.prop("tagName").toUpperCase() === "A";``` will to the trick.

nateabele:
Looks good. Feel free to submit those commits as a PR and I'll get them merged, thanks.

martinoss:
See this PR https://github.com/angular-ui/ui-router/pull/1668, thanks for merging.

nateabele:
Thanks for the fix.

PR #1668, opened 2015-01-03
fix($StateRefDirective): resolve support for svg anchor elements #1667

martinoss:
Fix the tagName comparism to make ui-sref also work inside an SVG.


nateabele:
Looks good. Please squash this down to one commit.

PR #1703, opened 2015-01-22
feat($uiViewScroll): change function to return promise

### LINKED ISSUE #1702: Return promise from uiViewScroll
I'm asking for the uiViewScroll function to return the timeout promise so that additional processing can occur once that is complete. The use case I have is with an iOS 7 bug on iPads that cause the screen to be blank until it's moved. In order to fix this I need to wait until the scroll completes and then move the screen an additional one pixel to make the content appear. For the time being I've had to use the source code from uiViewScroll and make my own function. Thanks!

Usage after issue resolved:
    $uiViewScroll(element).then(function() { $window.scrollTo(0,1);});

mikef438:
return the promise from the timeout for additional processing

Closes #1702

### DIFF
--- a/src/viewScroll.js
+++ b/src/viewScroll.js
@@ -42,7 +42,7 @@ function $ViewScrollProvider() {
     }
 
     return function ($element) {
-      $timeout(function () {
+      return $timeout(function () {
         $element[0].scrollIntoView();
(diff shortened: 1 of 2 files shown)


christopherthielen:
:+1: looks good

PR #1713, opened 2015-01-28
fix($urlMatcherFactory): regex params should respect case-sensitivity

### LINKED ISSUE #1671: Regex parameters do not respect case insensitivity
I tried the `$urlMatcherFactoryProvider` config:

```javascript
app.config(function ($urlMatcherFactoryProvider) {
  $urlMatcherFactoryProvider.caseInsensitive(true);
  $urlMatcherFactoryProvider.strictMode(false);
});
```
http://plnkr.co/edit/zdHlfVHBpnwDD3W5iUOW?p=preview

I traced this down to [this line](https://github.com/angular-ui/ui-router/blob/master/src/urlMatcherFactory.js#L122),  not sure if this is the intended behavior or it should be like this:

 ```javascript
type = $$UMFP.type(regexp || "string") || inherit($$UMFP.type("string"), { pattern: new RegExp(regexp, config.caseInsensitive ? 'i' : undefined) });
 ```

Thanks.

MohammadYounes:
Fixes #1671

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -10,7 +10,7 @@ var $$UMFP; // reference to $UrlMatcherFactoryProvider
  * of search parameters. Multiple search parameter names are separated by '&'. Search parameters
  * do not influence whether or not a URL is matched, but their values are passed through into
  * the matched parameters returned by {@link ui.router.util.type:UrlMatcher#methods_exec exec}.
- * 
+ *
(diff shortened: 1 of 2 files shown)


nateabele:
The patch looks good. Can you just amend the commit message to something like `fix($urlMatcherFactory): regex params should respect case-sensitivity` -- that way it'll work properly with our changelog generator. Thanks!

MohammadYounes:
@nateabele done!

nateabele:
Merged. Thanks again.

PR #1716, opened 2015-01-29
fix($urlMatcherFactory): unquote all dashes from array params

rolos79:
I noticed when using an array of uuids in a URL that the URL Matcher would only unquote the first dash it found. The fix was pretty straight forward and I have added tests.


christopherthielen:
Closes  #1717

PR #1718, opened 2015-01-30
feat(urlMatcherFactory): add support for dot notation in query parameters

drgould:
I have a use case for using dot notated query parameters (e.g. ?foo.bar=baz) and I noticed support was lacking from ui-router. As far as I can tell there is nothing that restricts dot notation in urls.

As part of this, it now recreates the implied structure for dot notated parameters. So "?foo.bar=baz" will result in $stateParams being: { "foo": { "bar": "baz" } }

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -26,6 +26,27 @@ function merge(dst) {
   return dst;
 }
 
+function copyParams(src, dest) {
+  var params = {};
+  forEach(src, function(param, key) {
+    var segments = key.split('.');
+    var tmp = params;
+    forEach(segments, function(segment, index) {
+      if (index < segments.length-1) {
+        if (!tmp[segment]) {
+          tmp[segment] = {};
+        }
+        if (!isObject(tmp[segment])) throw new Error("State parameter '" + key + "' is invalid: '" + segments.slice(0,index ).join('.') + "' is already defined.");
+        tmp = tmp[segment];
+      } else {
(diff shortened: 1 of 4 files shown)


nateabele:
Seems a little out there. Not sure how I feel about it. @christopherthielen?

Also, you committed release files.

drgould:
Sorry, forgot github automatically pulls in commits until the PR is merged. I took out the "release" commit.

christopherthielen:
I have no problem with supporting dots in param names, but the value being expanded into an object based on the param name feels really odd and unexpected to me.

I can't think of the use case that couldn't be solved in other ways. @drgould  what kind of params are you doing this with, and why expand the object out?

drgould:
I mainly need it for namespacing. We expose certain options that can be changed via query params. A few examples: `embedded.showNavigation` and `theme.headerFontColor`
It would be a lot easier if ui-router automatically created objects out of them so we could apply the options with a simple `$scope.opts.theme = $stateParams.theme` rather than each option separately (there are a lot).

nateabele:
> Maybe it could be something separate as to not drastically change the current function of $stateParams?

Yeah, I feel like maybe a mapping layer that sits between `$location.search()` and `UrlMatcher` would be more appropriate/flexible.

drgould:
I took out the object conversion so this can get merged in while I investigate a better way to the conversion.

PR #1726, opened 2015-02-03
feat(common): allow recursion in equalForKeys

flga:
**ui-sref-active** will not add the class if one of the passed parameters is an object literal because **common.equalForKeys** does not recurse:
http://plnkr.co/edit/2yfTHlhNPY1KWFQJ1pTY?p=preview

**Full disclosure:**
This is my first ever pr/contribution to OS. I read the guide and I think I'm on par with what's requested. If there's anything missing/incorrect please let me know.

Cheers

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -126,7 +126,13 @@ function equalForKeys(a, b, keys) {
 
   for (var i=0; i<keys.length; i++) {
     var k = keys[i];
-    if (a[k] != b[k]) return false; // Not '===', values aren't necessarily normalized
+    if(typeof a[k] === 'object' && typeof b[k] === 'object' ) {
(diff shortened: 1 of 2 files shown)


FabioAparicio:
Hi, what's the status on this?

christopherthielen:
@FabioAparicio I'm in the middle of a massive month-long crunch at my day job.  I haven't had time to look at much UI-Router stuff recently, sorry.  

I believe your idea behind the PR is sound though;  we should should be comparing params as objects, now that object params are supported. 

However, I don't think equalForKeys is the correct place to put that code.  I think we are moving away from that function towards using ParamSet.$$equals() to check if params are equal.  Second, we should probably be using angular.equals() to deeply compare objects, instead of coding a recursive call ourselves.

PR #1730, opened 2015-02-04
chore: remove jsdoc

### LINKED ISSUE #1729: jsdoc.js contains json, not javascript
config/jsdoc.js should be a .json file. It does not contain javascript, but json. This breaks javascript parsers like acorn who expect .js files to actually contain javascript.

Additionally, it seems like jsdoc is no longer used for the docs. It should probably be removed.

jimthedev:
Remove the jsdoc dependency since ng-doc is now being used

closes #1729

### DIFF
--- a/config/jsdoc.js
+++ b/config/jsdoc.js
@@ -1,6 +1,0 @@
-{
-  "plugins": [ "plugins/markdown" ],
-  "markdown": {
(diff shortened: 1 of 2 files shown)


jimthedev:
Fixes #1778.

PR #1736, opened 2015-02-05
feat($IncludedByStateFilter): add parameters to $IncludedByStateFilter

### LINKED ISSUE #1735: includedByState filter not fully working
`includedByState` improvement proposition

## tldr;

This works:

`<div ng-show="'**.foo.**' | includedByState">Work</div>`

and this don't.. (doesn't respect the second parameter, shown no matter what `id` is)

`<div ng-show="'**.foo.**' | includedByState:{id: 'new'}">Do not work</div>`

## Long version

The downside with it is that it doesn't use all the params from the `$state.includes` function, it only cares about the first one, which is `fullOrPartialStatename` (as reported in the [doc](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.filter:includedByState))

Here is a [plunker](http://plnkr.co/edit/tbvpxc8kkseoJOYBQXr4?p=preview) to show it live. In this example you can see that the red message is shown, no matter the params passed to the filter.

---

The issue come from [here in the code](https://github.com/angular-ui/ui-router/blob/master/src/stateFilters.js#L29) as it uses only a single param.

To resolve this, we can easily tweak the method to something like this: [Gist](https://gist.github.com/cyrilf/689dec37dd71c446e788), where `params` and `options` are given to the other method.

I may open a PR soon, but let me know **what you think about it?**

cyrilf:
Add parameters to the $IncludedByStateFilter:
- params
- options

This allows to use more parameters on this filter and not only the `fullOrPartialStatename` parameter.
Thanks to this improvement, the following now works (takes in account the id param):

`<div ng-show="'**.foo.**' | includedByState:{id: 'new'}">It works</div>`

Closes #1735

### DIFF
--- a/src/stateFilters.js
+++ b/src/stateFilters.js
@@ -27,8 +27,8 @@ function $IsStateFilter($state) {
  */
 $IncludedByStateFilter.$inject = ['$state'];
 function $IncludedByStateFilter($state) {
-  var includesFilter = function (state) {
-    return $state.includes(state);
+  var includesFilter = function (state, params, options) {
+    return $state.includes(state, params, options);
   };
   includesFilter.$stateful = true;
   return  includesFilter;
--- a/test/stateFiltersSpec.js
+++ b/test/stateFiltersSpec.js
@@ -25,7 +25,8 @@ describe('includedByState filter', function() {
     $stateProvider
       .state('a', { url: '/' })
       .state('a.b', { url: '/b' })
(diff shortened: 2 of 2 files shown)


cyrilf:
Any news about this PR? :wave:

nateabele:
Sorry for the wait.

cyrilf:
No problem.
Thanks! :sparkles:

Issue #733, opened 2014-01-04
Using Catch All parameters with ui-sref


Xiot:
I currently have a n-level deep route which can consist of a variable number of segments (ie directory browser).  I have the following state set up:

```
.state("folders", {
    url: "system/*folderPath",
    controller: "FolderController",
    templateUrl: "folders.html"
})
```

The transition is then navigated through an ui-sref
`<a ui-sref="folders({folderPath: "folder1/folder2"})" >Link</a>`

The ui-sref works fine and the location is changed to `system/folder1/folder2` however the link that shows up in the browser when you hover over it is `system/folder1%2Ffolder2` because the value is passed through `encodeURIComponent`  

What I am proposing is that for catch all route parameters, the value gets uri encoded, but should leave any forward slashes in, or at least have an option to make it work like that.

I can get around it by using ng-href to compose the uri, however then I lose the funtionality of ui-sref-active.

I did a quick and dirty hack of changing UrlMatcher.prototype.format to use

```
if (value != null) {
        if (this.sourcePath.indexOf("*" + params[i]) >= 0)
            result += value;
        else
            result += encodeURIComponent(value);
    }
```

but that only works for the shorthand `*parameterName` syntax.  The better way would probably be to keep a record of the catch-all parameters when the url format is parsed and use that to determine how to encode the value.

Let me know your thoughts

Hubrus:
Encountered a similar problem like @dmitry-dedukhin. Navigating to a state that has an url like: '/someAction/{id}/*params', makes two calls to the server for the template associated with the state.

AntJanus:
Just wanted to add that I have the same issue with the URLs. Does anyone have a hacky workaround until someone creates a core-modification to resolve this?

lopsided:
Ignore that, spoke too soon. It works as far as clicking the links in the browser works, but inspecting/copying the generated href shows it is still encoded. I still need a workaround like @Xiot suggested, although in keeping with the `any` type I used this:

`result += param.type.name == 'any' ? encoded : encodeURIComponent(encoded);`

Xiot:
I like the idea of basing it off of the type.
Maybe we would want a special 'path' type as well that would encode everything except for the slashes. That way we would get the benefits of both worlds

lopsided:
@Xiot I think this should do it. I've set up a new type which, on encoding, splits it by `/` and then in the format function still uses the `encodeURIComponent` for each path component and then joins it by `/` again.

[...]

PR #1738, opened 2015-02-06
feat(urlMatcherFactory): Added path param type to allow forward slashes in parameters, and not encode them

lopsided:
Solves #733

I couldn't see anywhere to add any documentation for the default types so I could add this to it. 

Please let me know if anything wants fixing, this is my first PR for angular.

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -342,7 +342,11 @@ UrlMatcher.prototype.format = function (values) {
       if (squash === false) {
         if (encoded != null) {
           if (isArray(encoded)) {
-            result += map(encoded, encodeDashes).join("-");
+            if (param.type.name == 'path') {
+              result += map(encoded, encodeURIComponent).join("/");
+            } else {
+              result += map(encoded, encodeDashes).join("-");
+            }
           } else {
             result += encodeURIComponent(encoded);
           }
@@ -619,6 +623,14 @@ function $UrlMatcherFactory() {
       is: angular.isObject,
       equals: angular.equals,
       pattern: /[^/]*/
+    },
+    path: {
+      encode: function(val) {
+        return val.split('/');
+      },
+      decode: angular.identity,
+      is: regexpMatches,
+      pattern: /.*/
     },
     any: { // does not encode/decode
       encode: angular.identity,
--- a/test/urlMatcherFactorySpec.js
+++ b/test/urlMatcherFactorySpec.js
@@ -530,6 +530,17 @@ describe("urlMatcherFactory", function () {
       expect(m.exec("/state/" + json1 + "/" + json2)).toEqual(params);
     });
 
+    it("should encode/decode path parameters without encoding the slashes", function () {
+      var m = new UrlMatcher("/{path:path}");
+      expect(m.exec("/some/path")).toEqual({ path: 'some/path' });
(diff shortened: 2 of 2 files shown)


eddiemonge:
How does it know when to stop with that param and move on to the next?

lopsided:
@eddiemonge it works in the same way to the "any" type, which also matches the same regex. I guess any parameter matching issues would also be encountered with that type, but this new type just doesn't encode the slashes

nateabele:
> Please let me know if anything wants fixing, this is my first PR for angular.

Thanks for the contribution. Please see http://prjs.radify.io/#/angular-ui/ui-router/pulls/1738

> I guess any parameter matching issues would also be encountered with that type, but this new type just doesn't encode the slashes

That makes it sound like it wouldn't work in a symmetrical way (also I don't see any tests for _de_-coding). Is that true?

ckniffen:
Wouldn't #2005 this fix the issue too?

lopsided:
@nateabele I've tried to fix the PR subject, I'm not sure why it's not gone all green though - am I missing something obvious?

You are right about the symmetry problems. The following test fails:

````
var m = new UrlMatcher("/{path:path}");
expect(m.exec("/some%20/path")).toEqual({ path: 'some /path' }); // comes out as "some%20/path"
````

But using the `any` type I found similar symmetry issues with the following test:

````
var m = new UrlMatcher("/{path:any}");
expect(m.exec("/some /path")).toEqual({ path: 'some /path' }); // ok
expect(m.format({ path: 'some /path' })).toBe('/some /path'); // comes out as "/some%20%2Fpath"
````

I hope you get what I'm trying to do here, and that I was trying to keep as close to the `any` type as possible. Though I do totally understand your concerns. 

I can actually fix the symmetry problem with my new `path` type by adding:

````
$normalize: function (val) {
            var parts = val.split('/');
          forEach(parts, function(part, key) {
              parts[key] = decodeURIComponent(part);
          });
          return parts.join('/');
        },
````

I tried something similar for the `any` type but it broke other tests. I haven't updated the PR with this since I'm not sure this is the right place to fix it. Any thoughts?

@ckniffen I don't think this is fixed in that other PR since that is about encoding and decoding of the slashes. I am trying to avoid encoding them at all.

nateabele:
> @nateabele I've tried to fix the PR subject, I'm not sure why it's not gone all green though - am I missing something obvious?

Sorry it's not obvious from the language (I'll fix that), but it's the commit message itself (not the PR title) that needs to be amended.

> But using the `any` type I found similar symmetry issues with the following test [...]

Well yeah, because `any` still follows the rules & regards itself as a normal path part.

oste:
I think a regex defined parameter should also not be encoded. Example ```url: "^{id:^(?!\/admin).*}",``` This is a catch all for any pattern except those that start with "/admin". So when it matches "my/forward/slash/pattern" it should just go ahead and match without redirecting by encoding the url.

nateabele:
> I think a regex defined parameter should also not be encoded.

Leaving aside the fact that this would be a serious and pervasive BC-break, it sort of ignores the fact that it breaks the boundary between `$location` and `URLMatcher`, since the encoding/decoding behavior is no longer consistent.

You're already able to implement a custom `URLMatcher` that does this.

PR #1755, opened 2015-02-15
coping with $injector:strictdi error

ErichBSchulz:
I was getting this error until I made this modification to my code:

"Uncaught Error: [$injector:modulerr] Failed to instantiate module ems8Ap due to: Error: [$injector:strictdi] function($stateProvider, $urlRouterProvider) is not using explicit annotation and cannot be invoked in strict mode"

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,15 @@ myApp.config(function($stateProvider, $urlRouterProvider) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


nateabele:
I don't think this is really necessary. Strict DI is an opt-in feature, and people should know what they're doing if they enable it. Thanks anyway for contributing.

PR #1777, opened 2015-02-25
use $templateRequest instead of $http directly

taylorcode:
In the angular core, $templateRequest is used instead of $http directly for fetching a template.

See ngRoute:

https://github.com/angular/angular.js/blob/master/src/ngRoute/route.js#L595

As well as ngInclude:

https://github.com/angular/angular.js/blob/3ae79c0105cbf58530dfb2d48865f25dcf2d3a2b/src/ng/directive/ngInclude.js#L229

$templateRequest internally calls $http with the $templateCache:

https://github.com/angular/angular.js/blob/f6272333127d908b19da23f9cd8a74052711795b/src/ng/templateRequest.js#L42

### DIFF
--- a/src/templateFactory.js
+++ b/src/templateFactory.js
@@ -9,8 +9,8 @@
  * @description
  * Service. Manages loading of templates.
  */
-$TemplateFactory.$inject = ['$http', '$templateCache', '$injector'];
-function $TemplateFactory(  $http,   $templateCache,   $injector) {
+$TemplateFactory.$inject = ['$http', '$templateCache', '$templateRequest', '$injector'];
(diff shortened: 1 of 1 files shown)


nateabele:
Couple issues:

 - Commit message not properly formatted (see contributor guide)
 - Tests fail
 - You left `$http` in the dependencies, even though it is now unused

Also, we support 1.2, and 1.2 doesn't support `$templateRequest`, so if anything, this PR would need to be held until sometime in the future.

wesleycho:
This PR is too thin, and has issues as mentioned prior - we should revisit this when it is more appropriate to (i.e. when UI Router drops active support for 1.2 and below).

Closing as a result.

PR #1779, opened 2015-02-25
Observe the ui-sref attribute so it can be updated if it changes.

rwlogel:
Currently if ui-sref changes after the directive is linked it is updated.  This change will observe the ui-sref and update the href if it changes.

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -84,7 +84,7 @@ function $StateRefDirective($state, $timeout) {
     restrict: 'A',
     require: ['?^uiSrefActive', '?^uiSrefActiveEq'],
     link: function(scope, element, attrs, uiSrefActive) {
-      var ref = parseStateRef(attrs.uiSref, $state.current.name);
+      var ref = null;
       var params = null, url = null, base = stateContext(element) || $state.$current;
       // SVGAElement does not use the href attribute, but rather the 'xlinkHref' attribute.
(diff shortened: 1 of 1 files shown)


nateabele:
The state name in `ui-sref` is a literal, not an expression.

rwlogel:
That isn't entirely true it is currently valid for the ui-sref to be:

```html
<select ng-model="foo">
    <!-- A selection box holding various entries that correspond to other states -->
</select>
<a  ui-sref="{{dynamicLinkToState()}}">Load Selected Entry</a>
```

This already works, but it will only reference the state provided at the time ui-sref is linked.  Adding the observe method will allow the user to dynamically change that entry.

nateabele:
That's incidental to the way Angular works, and is irrespective of `uiSref`'s design. There's an alternative proposal to support dynamic watching (and the attendant performance cost) in a separate directive.

plong0:
+1 for observing ui-sref attribute

@nateabele can I ask what is the alternative proposal around this?

nateabele:
@plong0 We just merged a `ui-state` directive to master. It'll be out in the next release.

plong0:
cool!  looking forward to the release.  would you say the master branch and ui-state on it are stable enough for production?  or better to wait for the official release?

nateabele:
All the tests are passing, so it's as stable as any other release. :smile:

PR #1798, opened 2015-03-05
Handle opening in new tab on command-click; fixes #1797

### LINKED ISSUE #1797: Open in new tab does not work on command-click of non-anchor element
When an element like a `<tr>` has a `ui-sref` attribute, <kbd>⌘ command</kbd> + <kbd>click</kbd> is ignored.

Example:

```html
<tr ng-repeat="item in items" ui-sref="view-item({itemId:item.id})">
```

andyfleming:
For non-anchor tags, command-click is ignored for opening in a new tab. This checks if the command key is down and uses window.open in combination with $state.href to properly open a ui-sref in a new tab on a non-anchor element.

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -134,7 +134,13 @@ function $StateRefDirective($state, $timeout) {
         if ( !(button > 1 || e.ctrlKey || e.metaKey || e.shiftKey || element.attr('target')) ) {
           // HACK: This is to allow ng-clicks to be processed before the transition is initiated:
           var transition = $timeout(function() {
-            $state.go(ref.state, params, options);
+            if (e.metaKey) {
+              options.absolute = true;
+              var url = $state.href(ref.state, params, options);
+              window.open(url,'_blank');
+            } else {
+              $state.go(ref.state, params, options);
+            }
           });
           e.preventDefault();
 


andyfleming:
@nateabele — Would you mind taking a look at this?

andyfleming:
or @christopherthielen? Would be much appeciated.

nateabele:
I don't understand. For non-anchor tags? Do browsers even have defined behavior for that?

andyfleming:
@nateabele It's not a browser behavior. I'm talking about when an element, that isn't an `<a>` tag, is made clickable to navigate somewhere.

Case 1:
When `<a ui-sref="example">` is clicked it navigates to the `example` route (as expected).

Case 2:
When `<tr ui-sref="example">` is clicked it navigates to the `example` route (as expected).

Case 3:
When `<a ui-sref="example">` is <kbd>⌘ command</kbd>-clicked it opens the `example` route in a new tab (as expected).

Case 4:
When `<tr ui-sref="example">` is <kbd>⌘ command</kbd>-clicked it navigates to the `example` route in the same window and **NOT in a new tab as expected**.

nateabele:
Yeah, this is out-of-scope, sorry. The fact that you can add `ui-sref` to a `<tr />` tag and it actually does something is incidental, not explicitly-defined behavior.

We're definitely not going out of our way to re-implement things that browsers don't even do. That should be a separate directive outside of UI-Router.

PR #1804, opened 2015-03-06
uiSrefActive and uiSrefActiveEq on same element

hiersekornc:
uiSrefActive and uiSrefActiveEq should now work as expected when used on the same element.
Direct matching states will be given the uiSrefActiveEq's class in addition to uiSrefActive's class.
Useful for differently coloring direct states vs included state.


nateabele:
Hm, looks like either it broke some tests, or the tests need to be updated. Also, the change should have it's own accompanying tests to prove the behavior.

PR #1809, opened 2015-03-09
feat($state): added 'state' to state reload method (feat no.1612)

maoryosef:
- modiefied options.reload parameter in transitionTo method to accept the
   following:
   {reload : true} will reload all states
   {reload : 'foo.bar'} will reload foo.bar, and any children
   {reload : stateObj} will reload state found in stateObj, and any
   children.
 - added 'state' parameter(string|object) to $state.reload method,
   to be passed to $state.transitionTo method as the reload option.
 - test the $state.reload with state parameter

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -820,11 +820,33 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
      * });
      * </pre>
      *
+     * @param {string=|object=} state - A state name or a state object, which is the root of the resolves to be re-resolved.
+     * @example
+     * <pre>
+     * //assuming app application consists of 3 states: 'contacts', 'contacts.detail', 'contacts.detail.item' 
+     * //and current state is 'contacts.detail.item'
+     * var app angular.module('app', ['ui.router']);
+     *
+     * app.controller('ctrl', function ($scope, $state) {
+     *   $scope.reload = function(){
+     *     //will reload 'contact.detail' and 'contact.detail.item' states
+     *     $state.reload('contact.detail');
+     *   }
+     * });
+     * </pre>
+     *
+     * `reload()` is just an alias for:
+     * <pre>
+     * $state.transitionTo($state.current, $stateParams, { 
+     *   reload: true, inherit: false, notify: true
+     * });
+     * </pre>
+
      * @returns {promise} A promise representing the state of the new transition. See
      * {@link ui.router.state.$state#methods_go $state.go}.
      */
-    $state.reload = function reload() {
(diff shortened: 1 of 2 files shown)


christopherthielen:
Thanks, really appreciate the work.  

If @nateabele approves of this API change, I'll want to merge this in manually, tweaking lines 1002-1020 to closer match the ui-router coding style.

src/state.js
```diff
@@ -975,13 +999,25 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
...
+        if (!isDefined(findState(stateName))) {
+          throw new Error("No such state '" + stateName + "'");
+        }
+
+        while (state && state === fromPath[keep] && state.toString().toLowerCase() !== stateName.toLowerCase()) {
+          locals = toLocals[keep] = state.locals;
+          keep++;
+          state = toPath[keep];
+        }
       }
```
christopherthielen: The block above should be something closer to:

```
var reloadState = options.reload === true ? fromPath[0] : findState(options.state);
 if (options.reload && !reloadState) {
          throw new Error("No such reload state '" + options.reload + "'");
        }
var skipTriggerReloadCheck = !!reloadState;
        while (state && state === fromPath[keep] && state !== reloadState) {
          locals = toLocals[keep] = state.locals;
          keep++;
          state = toPath[keep];
        }

```

oriweingart:
Done, refactor the code according to your comment.

nateabele:
:+1: Just squash the commits.

maoryosef:
Commits were squashed

src/state.js
nateabele: @christopherthielen Do you want this fixed before we merge?

nateabele: Or, wait looks like it was already fixed?

christopherthielen: yeah, they already made the changes I asked for

lefterisk:
Good work people. @nateabele , someone should add this in the documentation, it took me half a day looking into workarounds for this problem only to find that its solved already, just not mentioned in the documentation. (This is a common problem apparently and people are using some scary workarounds)

nateabele:
>  @nateabele , someone should add this in the documentation

@lefterisk Go for it. :smile:

lefterisk:
@nateabele created pull request #2347 for this :)

nagyg-bvictor:
I'm not sure if I understand this right, but the reload string must be an absolute state name? I want to pass a relative state path like '.child'. I think it would be more straightforward.

PR #1838, opened 2015-03-25
Fix critical bug - routing stop working in some cases

jamm:
I don't know how it's possible, but in all browsers I see exception with message about .params of undefined.
This is a small, simple and safe fix.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -96,7 +96,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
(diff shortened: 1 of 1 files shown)


nateabele:
Unit test?

jamm:
@nateabele sorry, have no time to understand your testing rules. It's 1-line fix, think it will not be difficult to write simple test for somebody who already familiar with your tests.

nateabele:
>  It's 1-line fix

Running the tests is a 1-line command.

It's a 1-line fix that I have no idea why you want added to the codebase I'm responsible for maintaining.

Whatever your issue is, apparently you're the only one who has it.

jamm:
@nateabele you could write test faster, then writing so supercilious response. I told you already, that reading codebase of your tests will take more time than writing any test. Don't want - don't eat. Will use forked version.

Issue #1525, opened 2014-11-11
preventDefault() on back button $stateChangeStart causes history to be overwritten


AngusHelm:
Plunker: http://plnkr.co/edit/CmuEPKjRv0KMJYywm2wQ

I've got a linear series of states, and I'm trying to popup a confirmation dialog before the user moves backwards at a certain step with preventDefault() on $stateChangeStart. Maybe browsers don't allow this, but the preventDefault not only doesn't work but modifies the history stack so that what was a linear step1->step2->step3 becomes step1->step3->step3.

mgrahamjo:
This is related to the issue I opened today: https://github.com/angular-ui/ui-router/issues/1780

You're seeing this behavior, as I am, because ui-router checks to see if default was prevented after the redirect happens instead of before (that's where the new history entry is coming from). If they see that you called preventDefault, then they cancel the transition and call $location.url(location) to send you back to the starting location.

The fix is to check for preventDefault before executing the redirect during transitionTo.

arcanis:
That's something that I really would like to see solved. It makes impossible to have asynchronous state validation without breaking the back history (so it doesn't work with [angular-permission](), for example).

I've think about how to workaround it, but I can't see any real way. It would be great if the core allowed it.

mgrahamjo:
The fix for my purposes was to add in a $beforeStateChange event and listen for that instead of $stateChangeStart. I have this in the run() block of app.js:

    var transitionTo = $state.transitionTo;
    $state.transitionTo = function(to, toParams, options) {
        var from = $state.$current,
            fromParams = $state.params;

        to = to.name ? to : $state.get(to);

        if (options.notify !== false && $rootScope.$broadcast('$beforeStateChange', to, toParams, from, fromParams).defaultPrevented) {
            return $q.reject(new Error('transition prevented'));
        } else {
            return transitionTo(to, toParams, options);
        }
    };

txikito:
Hi,

New to this group, sorry for any inconvenience...

mgrahamjo: tried your solution in a simple example, and could not get it working, it still breaks when navigating with the back button and preventing that navigation in the beforeStateChange event. Please, see this plunker to observe strange behaviour in the history stack: 

http://embed.plnkr.co/M8IL5v/

If you navigate from index to page 1, then page 2, and try to naviagate back with explorer back button, you'll see that even though the state change is prevented, the change in the history stack still occurs (at least apparently).

[...]

PR #1845, opened 2015-03-28
feat($state): add $stateChangeValidation event

arcanis:
## Why?

> It [is] impossible to have asynchronous state validation without breaking the back history (so it doesn't work with angular-permission, for example).

([full comment](https://github.com/angular-ui/ui-router/issues/1525#issuecomment-87092724))

Until now, if we wanted to pause the state evaluation and resume it later, we had to do something like this :

```js
$rootScope.$on('$stateChangeStart', function (e) {
  e.preventDefault();
  doSomeLongRunningTask().then(function () {
    $state.go( 'foo' );
  });
});
```

However, it had a major drawback : it broke the backward navigation (because the preventDefault() / $state.go combo was rewriting the history, so that if you had a A->B->C history, then pressing the back button would bring you to B, then pressing it again would bring you to C instead of A).

## How?

This PR adds a $stateChangeValidation right before emitting the start event. At this point, listeners are still able to "hold" the transition execution by calling a function given to them as the last parameter of the event. Multiple listeners may hold the execution, and a single listener may also hold multiple time.

Each validator is a value, which may be a promise (in such case, we gently wait for it to be resolved). If at least one of the validator is falsy, the transition is cancelled. No more, no less.

The $stateChangeValidation event may be preventDefault()'d, to bypass any validation.

## Example

```js
$rootScope.$on('$stateChangeValidation',
function(event, toState, toParams, fromState, fromParams, registerValidator){
  registerValidator(soAsyncStore.get('userUid'));
  ...
})
```

## Backward Compatibility

This PR shouldn't break anything. If no listener ask for a hold on the transition, then the code will follow the current codepath, and everything will go as usual (we won't even wait for a dummy promise to resolve).

I've added tests which should prevent the feature from breaking in the future.

## Why not $locationChangeStart?

Because $locationChangeStart doesn't yet know what state will be navigated to. $stateChangeValidation knows, and can use this information to tune its permissions.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -141,9 +141,9 @@ module.exports = function (grunt) {
   grunt.registerTask('widedocs', 'Convert to bootstrap container-fluid', function () {
     promising(this,
       system(
-      'sed -i.bak ' + 
-      '-e \'s/class="row"/class="row-fluid"/\' ' + 
-      '-e \'s/icon-cog"><\\/i>/icon-cog"><\\/i>Provider/\' ' + 
+      'sed -i.bak ' +
+      '-e \'s/class="row"/class="row-fluid"/\' ' +
+      '-e \'s/icon-cog"><\\/i>/icon-cog"><\\/i>Provider/\' ' +
       '-e \'s/role="main" class="container"/role="main" class="container-fluid"/\' site/index.html')
     );
   });
--- a/src/state.js
+++ b/src/state.js
@@ -122,7 +122,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
     if (path) {
       if (!base) throw new Error("No reference point given for path '"  + name + "'");
       base = findState(base);
-      
+
       var rel = name.split("."), i = 0, pathLength = rel.length, current = base;
 
       for (; i < pathLength; i++) {
@@ -257,9 +257,9 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
    * @methodOf ui.router.state.$stateProvider
    *
    * @description
-   * Allows you to extend (carefully) or override (at your own peril) the 
-   * `stateBuilder` object used internally by `$stateProvider`. This can be used 
-   * to add custom functionality to ui-router, for example inferring templateUrl 
+   * Allows you to extend (carefully) or override (at your own peril) the
+   * `stateBuilder` object used internally by `$stateProvider`. This can be used
+   * to add custom functionality to ui-router, for example inferring templateUrl
    * based on the state name.
    *
    * When passing only a name, it returns the current (original or decorated) builder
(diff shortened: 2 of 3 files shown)


arcanis:
I'm rebasing to make the merge easier, should be ready in a few minutes. Just wanted to get feedback! :)

nateabele:
Hi @arcanis, thanks for your work on this. Clearly the problem you state is something that needs to be addressed. However, your change is somewhat invasive and adds quite a bit of code. I'm looking to see if this can be addressed in a simpler and more generalized way by decreasing coupling between components.

In the meantime, would you mind submitting the whitespace fixes as a separate PR? That would at least make it easier to read what this one is doing. Thanks.

arcanis:
I'll do that. By itself, this PR doesn't add so much code (only 50, not counting whitespace changes required when indenting a part of the code).

nateabele:
Yeah, I just noticed the link you posted. The actual change is pretty minimal (mostly docs) after all. However, as much as I'd like to just merge this, the problem you've identified falls within a pattern that's repeated itself to nontrivially increasing complexity. Rather than continuing down the path, I'd rather go back and fix the design, even if it means prolonging the short-term pain. Hope that makes sense.

I'll be working on some related things this week, so I've flagged this for myself, so that this use case will be considered in the design changes.

arcanis:
For what it worth, I'm using this PR in production, and it worked flawlessly until now. Hope you can find another solution for the 1.0 release, but it still think it might be interesting for the 0.X branch :)

nateabele:
Yeah, you know, I've changed my mind on this. If you want to rebase and resubmit per the above, I'll merge this.

christopherthielen:
It should be noted that we are likely deprecating the state events in 1.0

arcanis:
@nateabele Sure, here it is, freshly rebased.

@christopherthielen What will be the 1.0 solution to this use case?

christopherthielen:
@arcanis in 1.0, each step of a transition may add resolves to the transition.  You can register a global transition hook, check to-state criteria, and add a resolve to the transition if it should have one.

Something along these lines:
```
$transition.on({to: "*", from: "*"}, function($transition$) {
  var state = $transition$.to.state();
  if (state.requiresAuth) {
    return { 
      authentication: function(authService) { return authService.checkServerSide(); }
    }
  }
});
```

alxndrsn:
What was the final verdict on this PR?  Any chance of it getting merged?  Is there an alternative approach to fixing the back behaviour?

nateabele:
I'm still willing to merge the feature, but this PR is still a mess...

arcanis:
Well yeah, it doesn't merge anymore, since it hasn't been merged the last time I rebased it, as per your request .. :(

@alxndrsn You might want to try the [1.0 beta branch](https://github.com/angular-ui/ui-router/tree/feature-1.0), which apparently implements a new way to asynchronously validate a route transition.

nateabele:
> In the meantime, would you mind submitting the whitespace fixes as a separate PR? That would at least make it easier to read what this one is doing. Thanks.

:smile:

alxndrsn:
@nateabele I've submitted a rebased PR
@arcanis I might give 1.0 a go at some point, but not keen right now for a major version change.  Thanks for getting this patched in the first place though :+1:

txikito:
Hi @arcanis,

I tried to manually add your PR to the main angular-ui-router file, but I could not get a satisfactory result. Please, take a look at the following plunkr to demonstrate the issue, if you have some time:
http://embed.plnkr.co/b0euqt/

If you navigate from main to Page1, then to Page2, and try to navigate back with the back button, the event is actually prevented, but a strange effect is seen in the history stack, as the forward button gets activated, when it should not, as the state has not been changed.

Any help with this issue will be welcome.

Thank you,

alxndrsn:
@txikito you might have some luck with my update to @arcanis's work at https://github.com/angular-ui/ui-router/pull/2328

txikito:
Hi @alxndrsn,

Thanks for the help. I tried your suggestion with no luck. 

Finally, I managed to get the issue "working", with a combination of using @mgrahamjo's  beforeStateChange solution, and in the back button case, attaching to $locationChangeStart and preventing the event from there.

I am aware that it is not by far the best solution, but it seems to be working for me in all the cases I am facing.

Thank you.

christopherthielen:
Thank you for your PR!  

As I mentioned in the rebased PR #2328, I don't think this approach aligns with our vision for ui-router's future.  I think the transition hooks in 1.0 fully address the async validation use case, and we have deprecated (and removed by default) the state events.  

angular-permission was mentioned as a driver; I see that there has been some activity already migrating to the 1.0 transition hooks: https://github.com/Narzerus/angular-permission/issues/81#issuecomment-170543911

Hopefully this PR or #2328 addresses your use cases for now.  Please take a look at the 1.0 alpha release and let me know if there is any reason the transition hooks are not going to work for you.

PR #1455, opened 2014-10-15
feat(ui-sref): support URL fragments in html5Mode

encounter:
This provides support for specifying URL fragments in `ui-sref` attributes when HTML5Mode is enabled.

Thanks to @mismith for his original work in #1187.

PR #1855, opened 2015-04-02
feat($state): support URLs with #fragments in html5Mode

mismith:
Allow setting the `#` param to update the location fragment; e.g.: `ui-sref="page({name: 'name', '#': 'frag'})"` or `$state.go('page', {name: 'name', '#': 'frag'})`

This is hoping to solve the various issues found in: #701, #1187, and #1455

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -820,11 +820,33 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
      * });
      * </pre>
      *
+     * @param {string=|object=} state - A state name or a state object, which is the root of the resolves to be re-resolved.
+     * @example
+     * <pre>
+     * //assuming app application consists of 3 states: 'contacts', 'contacts.detail', 'contacts.detail.item' 
+     * //and current state is 'contacts.detail.item'
+     * var app angular.module('app', ['ui.router']);
+     *
+     * app.controller('ctrl', function ($scope, $state) {
+     *   $scope.reload = function(){
+     *     //will reload 'contact.detail' and 'contact.detail.item' states
+     *     $state.reload('contact.detail');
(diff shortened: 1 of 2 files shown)


nateabele:
Looks pretty good, thanks for putting this together. Just to be clear, what's the behavior in non-HTML5-mode? Do you just not get a fragment?

mismith:
I would assume so, I haven't really thought that through. What would be best? Do browsers even let you set a URL like `#!/hello/world#frag`

nateabele:
Yeah, you can put whatever you want in the URL. I believe the `$location` service takes care of parsing for you. See here: https://docs.angularjs.org/api/ng/service/$location#hash

If you can amend the commit with a test that verifies that this change mimics the above behavior (you might have to break off the `'#'` and pass it to `$location.hash()` after the call to `$location.url()`), then I'll merge this. Thanks a lot for your work on this.

mismith:
So which would you prefer, that I enable this behaviour for non-html5Mode as well, or test that it has no effect when in that mode?

nateabele:
I think the effect should be the same as using `$location.hash()`, so, you get an extra `#` in the URL, that the `$location` service parses out transparently. Does that make sense?

nateabele:
Again, I _think_ you could probably just Make It Work™ by passing the `params['#']` to `$location.hash()` instead of appending it to `url`.

mismith:
I think the issue with that would be the order of operations? If I used `.hash()` in place of appending the the url, it would fire right away, no? (rather than after the state changes)

nateabele:
Not sure, I just assumed it would work. Anyway, if the tests pass in the new PR in both modes, good enough for me.

PR #1869, opened 2015-04-09
fix(uiSref): use Object's toString instead of Window's toString

dtritus:
Fix 'Invalid Calling Object' error in IE11


nateabele:
Thanks for the fix.

PR #1877, opened 2015-04-13
feat(uiSref): add ui-sref-on to expose transition promise

### LINKED ISSUE #1863: RFC: Add a ui-sref-resolve directive similar to ui-sref-active(-eq)
The purpose of this is a loading state icon on navigation elements while fetching all data through resolve.

If a user clicks a button in the navigation to change the state (to: `app.user.profile`) the app needs to fetch the profile data first (`resolve: { profile: function($http) { $http.get... } }`). During the latency of the network request I'd like to have the button pressed (`<a class="btn btn-primary">`) to get an extra class (`"btn-loading"`) to show the user an indicator that the click was registered.

The way this currently has to be done:

``` html
<nav>
  <a class="btn btn-primary" ng-class="{'btn-loading': changingState}" ui-sref="app.user.profile">Profile</a>
</nav>
```

``` javascript
function Controller($scope, $state) {
  $scope.changingState = false; // simplified here would need one for each button
  $scope.go = function(state) {
    $scope.changingState = true;
    $state.go(state).then(function() { $scope.changeingState = false; })
  }
}
```

Proposed change:

``` html
<nav>
  <a class="btn btn-primary" ui-sref-resolve="btn-loading" ui-sref="app.user.profile">Profile</a>
</nav>
```

lordnox:
Fixes #1863

@see https://github.com/angular-ui/ui-router/issues/1863

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -134,7 +134,12 @@ function $StateRefDirective($state, $timeout) {
         if ( !(button > 1 || e.ctrlKey || e.metaKey || e.shiftKey || element.attr('target')) ) {
           // HACK: This is to allow ng-clicks to be processed before the transition is initiated:
           var transition = $timeout(function() {
-            $state.go(ref.state, params, options);
+            var $transition = $state.go(ref.state, params, options);
+            if($transition && attrs.uiSrefOn) {
+              scope.$eval(attrs.uiSrefOn, {
+                $transition: $transition
+              });
+            }
           });
           e.preventDefault();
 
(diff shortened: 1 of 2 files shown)


nateabele:
Looks like one of the tests failed?

lordnox:
Hm.. could it be that the return value of `$state.go` differs when run on travis or it might be just the `karma:past` test I missed...
Should it not return a promise? *sigh*

The only solution I can think of is to wrap the returned value in a `$q.when()`...
here: https://github.com/raynode/ui-router/blob/master/src/stateDirectives.js#L140

nateabele:
> The only solution I can think of is to wrap the returned value in a $q.when()...
here: https://github.com/raynode/ui-router/blob/master/src/stateDirectives.js#L140

Yeah, I think that's basically what the rest of the tests do. You could also check out the `$resolve` tests, which have a couple wrappers for testing promises. Thanks. Let me know if you need any help.

nateabele:
Closing in favor of #1986.

PR #1884, opened 2015-04-14
fix(uiSrefActive): Apply active classes on lazy loaded states

### LINKED ISSUE #915: Make ui-sref-active lazy load friendly
I've been doing some work with making my an app completely dynamic and I've found that the internal stack of `states`, though updated upon dynamically loaded states, does not have its changes reflected in the uiSrefActive directive.

In the [$stateChangeSuccess](https://github.com/angular-ui/ui-router/blob/91a568454cc600aa4f34dedc8a80c9be1130b1c5/src/stateDirectives.js#L209) update function, by simply adding the following code, it all instantly stated to work;

```js
function update() {

    var ref;

    if (!state) {
        ref = parseStateRef($attrs.uiSref);
        state = $state.get(ref.state, stateContext($element));
    }

    if ($state.$current.self === state && matchesParams()) {
    ...
}
```

Cheers guys and keep up the good work!

kasperlewau:
#### Why? 
When generating new $state definitions through lazy loaded content, `uiSrefActive` will not reflect class changes on the associated anchor tag as the $state definition was not available in the initial stack. 

---

#### How? 
By pushing a *dummy* state definition to the initial stack of states found on application initialization with the `lazy: true` flag, class reflection onto the view is now working. However, this does not hold true (entirely) when launching your application **on** a lazy loaded state. 

Take the following structure: 

```js
$stateProvider.state('home', { url: '/' });
$stateProvider.state('home.lazy1', { url: '/l1' }); // State definition is added in a module loaded during runtime.
$stateProvider.state('home.lazy2', { url: '/l2' }); // State definition is added in a module loaded during runtime.
```

```html
<a ui-sref="home" ui-sref-active="active">Home</a>
<a ui-sref="home.lazy1" ui-sref-active="active">Lazy 1</a>
<a ui-sref="home.lazy2" ui-sref-active="active">Lazy 2</a>
```

If we were to launch the application at `localhost:9000/l1`, the `home.lazy1` anchor tag would get the `active` class applied to it. However, when navigating to `/l2`, `home.lazy2` would **not** get the `active` class applied to it. 

The same holds true for: 
* Launching on `/l2` and navigating to `/l1`. 
* Launching on `/` and navigating to either of the lazy loaded states.

---

[jsBin w/fix (see the bottom)](http://jsbin.com/fememinina/5/edit?html,js,output)
[gist of possible setup](https://gist.github.com/kasperlewau/4b3afa5587a55ccd7481)

My first PR to ui-router, feel free to dissect the commit/PR, here's hoping you won't have too much to do!

Fixes #915

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -239,13 +239,13 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
       // Allow uiSref to communicate with uiSrefActive[Equals]
       this.$$addStateInfo = function (newState, newParams) {
         var state = $state.get(newState, stateContext($element));
-        if (state) {
-          states.push({
-            state: state,
-            params: newParams
-          });
-          update();
-        }
+
+        states.push({
+          state: state || { name: newState },
+          params: newParams
+        });
+
+        update();
       };
 
       $scope.$on('$stateChangeSuccess', update);
--- a/test/stateDirectivesSpec.js
+++ b/test/stateDirectivesSpec.js
@@ -392,11 +392,12 @@ describe('uiStateRef', function() {
 });
 
 describe('uiSrefActive', function() {
-    var el, template, scope, document;
+  var el, template, scope, document, _stateProvider;
 
   beforeEach(module('ui.router'));
 
   beforeEach(module(function($stateProvider) {
+    _stateProvider = $stateProvider;
     $stateProvider.state('top', {
       url: ''
     }).state('contacts', {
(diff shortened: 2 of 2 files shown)


src/stateDirectives.js
```diff
@@ -244,8 +244,15 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
             state: state,
             params: newParams
           });
-          update();
```
nateabele: Better:

```javascript
states.push({
  state: state || { name: newState },
  params: newParams,
  lazy: !!state
}); 
```

No `if` block.

kasperlewau: Agreed, I'll amend the commit in a wee bit.

nateabele:
Made one refactoring comment. If you can fix that & squash your commits, it's :+1:.

Nice work, especially for your first PR. Thanks again for the fix.

src/stateDirectives.js
kasperlewau: Hm.

* A single `!` - `if state is not predefined, set lazy to true`
* Triple `!!!` - `if state is not predefined, set lazy to true - and cast to bool`

Double `!!` would reverse the desired effect if the state was present in the initial stack, no?

It's getting late, my mind isn't working quite right - but I'm seeing a single failed spec example with the suggested changes (`it('should match on any child state refs`), none of the newly added specs are failing though. I'll go get some fresh air and come back in a bit.

kasperlewau: There, fresh air. 

We're looking at using either a single or a triple exclamation mark here to reach the desired effect - with a non-negating cast to boolean we're effectively setting lazy to true if the state was defined on initial load of the application, which is the reverse of what we want to do. 

So I'd say it's a matter of preference/taste now really; Single or Triple? I'm inclined to say triple for readability. 

poke @nateabele

kasperlewau:
Alrighty - time for bed. I managed to sneak some more refactoring in there, effectively removing the `lazy` flag. Simply pushing a `{ name: newState }` dummy object to the initial stack of states take care of the problem(s). 

I'll leave it at that for the time being and give it all a whirl in my sample application tomorrow morning. 

(and the petition of ! vs !!! is now moot :smile:).

kasperlewau:
My only concern now is when ui-sref-active is used in conjunction with a ui-sref directive that's targeting a nonexistent state that *wont ever* be lazily loaded. 

I'm not sure how one would detect that, as it's somewhat the nature of the lazy beast.

src/stateDirectives.js
nateabele: Yeah, agreed, good catch. Triple seems good to me.

nateabele:
Never mind my other comment. Finally finished reading the thread/updated code. Well-played.

nateabele:
> I'm not sure how one would detect that, as it's somewhat the nature of the lazy beast.

Yeah, exactly. There's only so much we can do it.

PR #1891, opened 2015-04-17
feat(uiView): support set scroll to top before transition

bensechrist:
Allow setting `resetscroll` as an attribute of `uiView` to
set the scrollTo of the window to 0 0 before transitioning
to a new state. Before the window was not reset and the
scroll on the new state would be the scroll of the previous
state.

### DIFF
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -26,6 +26,9 @@
  * functionality, call `$uiViewScrollProvider.useAnchorScroll()`.*
  *
  * @param {string=} onload Expression to evaluate whenever the view updates.
+ *
+ * @param {string=} resetscroll It allows you to reset the window scroll to 
+ * the top before transitioning to a new state.
  * 
  * @example
  * A view can be unnamed or named. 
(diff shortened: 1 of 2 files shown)


nateabele:
I'm not opposed to this kind of functionality, but at a minimum, it would need to be a patch against [`$uiViewScrollProvider`](https://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$uiViewScrollProvider), and ideally would be something parameterized that allowed you to return an arbitrary value, rather than being hard-coded to (0, 0).

bensechrist:
So using the $uiViewScroll function to scroll the window before destroying the current view?

nateabele:
On second thought, let me think about it. I don't want `uiView` muddied by entangling it with decisions about scroll behavior (since that logically belong elsewhere), but maybe there's an opportunity here for a more flexible API that allows people to design the behavior they want, instead of being use-case-specific, as per this patch.

bensechrist:
Agreed. Came up with this today to avoid the odd user interactions of changing states and ending up at the bottom/middle of the page or the page oddly scrolling to the top.

nateabele:
Yeah, the available options are less than amazing for all use cases. We just built a few additional things off of what's available in Angular core.

kasperlewau:
Back when autoscrolling was on by default (and I've never quite liked *being scrolled*), I simply did the following to nullify the default behaviour.

```js
$provide.decorator('$uiViewScroll', function () {
  return angular.noop;
});
```

This could easily be rewritten to allow for full on control of the scrolling behaviour; 

```js
$provide.decorator('$uiViewScroll', function ($delegate) {
  return function (uiViewElement) {
    /** roll your own **/
  }
});
```

Not sure if that's at all what you had in mind - but I could see a decorator being 'masked' by a method on the $uiViewScrollProvider to roll your own solution. 

```js
function myCustomScroll (uiViewEl) {
  /* do some cool scrolly stuff */
}

$uiViewScrollProvider.overrideWith(myCustomScroll);
```

But perhaps that's letting loose of the reigns too much? Perhaps a slightly more configuration based API is desirable for consumer modification(s)?

nateabele:
> Not sure if that's at all what you had in mind

Not quite. It's partly a timing issue: he wants to reset the scrollbar _before_ the transition happens.

kasperlewau:
Ah I see. Then it's a wee bit more involved than overriding the $uiViewScroll functionality.

PR #1921, opened 2015-04-27
New urlMatcher param type : optionalint

### LINKED ISSUE #1920: There should be an "optional" int parameter type
The new "int" type is great but it is now mandatory. There should be a parameter type that match "/200/" and also "//".

court-jus:
Allows state to be defined with an optional parameter that is returned
as int if specified.

Fixes #1920


nateabele:
You're complecting generalized behavior with a data type.

court-jus:
I'm sorry, I don't understand your comment, I've fixed my dev and now it works for me but I'll make it "ok for integration" if I know what I have to do

court-jus:
I've found a cleaner solution and explained it in #1920

Issue #900, opened 2014-02-18
Make ui-sref accept objects and not only Strings


mgonto:
Hey,

Sometimes we actually want to link to another state based on a value on scope and not only by text.

I think it might be a good idea to make ui-sref accept values and call `$eval` on it. I created my own directive for this. You may want to use something similar. I can PR this as well to your code if you guys want.

https://github.com/mgonto/mgo-ui-router-goto/blob/master/uiGoto.js

Thanks!

christopherthielen:
Fixed in 3831af1dc71b601351e6694af0665a77297f8f7f

PR #1932, opened 2015-05-02
feat(uiState): add ui-state directive

kasperlewau:
ui-sref with dynamic state/option/param pointers.

Usage: 

```html
<a ui-state="state" ui-state-params="params" ui-state-opts="opts">
  Link
</a>
```

```js
$scope.state = 'myStateName';
$scope.params = { id: 5 };
$scope.opts = { reload: true };
```

The resulting `href` would be pointed to `myStateName({ id: 5 })` and `$state.go` would be called with `reload: true`. 

--- 

It's also possible to specify the params straight into the `ui-state` directive as such:

```html
<a ui-state="stateWithParam"></a>
```

```js
$scope.stateWithParam = 'myStateName({ id: 10 })';
```

--- 

I started off writing this as a completely standalone directive, but given that it is desirable for it to work much like `ui-sref`, I felt there was too much duplication going on. `ui-state` can therefore be seen as a wrapper around the behaviour of `ui-sref`, nothing more to it really.

Now, this thing is **heavy** on the `$eval` and slightly so on the `$watch` side of things, so if someone has any nifty ideas on how to make that performant (I have not tested the performance aspects of ui-state in depth), that'd be super sweet. 

--- 

Do let me know if you think there's a test missing among the added ones - I can't say I'm 100% certain that I've tested every aspect of a 'dynamic ui-sref directive'. 

Reference issues: #395 #900
[Suggested implementation](https://github.com/angular-ui/ui-router/issues/395#issuecomment-56884667)

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -24,17 +24,17 @@ function stateContext(el) {
  * @restrict A
  *
  * @description
- * A directive that binds a link (`<a>` tag) to a state. If the state has an associated 
- * URL, the directive will automatically generate & update the `href` attribute via 
- * the {@link ui.router.state.$state#methods_href $state.href()} method. Clicking 
- * the link will trigger a state transition with optional parameters. 
+ * A directive that binds a link (`<a>` tag) to a state. If the state has an associated
+ * URL, the directive will automatically generate & update the `href` attribute via
+ * the {@link ui.router.state.$state#methods_href $state.href()} method. Clicking
+ * the link will trigger a state transition with optional parameters.
  *
- * Also middle-clicking, right-clicking, and ctrl-clicking on the link will be 
+ * Also middle-clicking, right-clicking, and ctrl-clicking on the link will be
  * handled natively by the browser.
  *
- * You can also use relative state paths within ui-sref, just like the relative 
+ * You can also use relative state paths within ui-sref, just like the relative
  * paths passed to `$state.go()`. You just need to be aware that the path is relative
- * to the state that the link lives in, in other words the state that loaded the 
+ * to the state that the link lives in, in other words the state that loaded the
  * template containing the link.
  *
  * You can specify options to pass to {@link ui.router.state.$state#go $state.go()}
@@ -42,22 +42,22 @@ function stateContext(el) {
  * and `reload`.
  *
  * @example
- * Here's an example of how you'd use ui-sref and how it would compile. If you have the 
+ * Here's an example of how you'd use ui-sref and how it would compile. If you have the
  * following template:
  * <pre>
  * <a ui-sref="home">Home</a> | <a ui-sref="about">About</a> | <a ui-sref="{page: 2}">Next page</a>
- * 
+ *
  * <ul>
  *     <li ng-repeat="contact in contacts">
  *         <a ui-sref="contacts.detail({ id: contact.id })">{{ contact.name }}</a>
  *     </li>
  * </ul>
  * </pre>
- * 
+ *
  * Then the compiled html would be (assuming Html5Mode is off and current state is contacts):
  * <pre>
  * <a href="#/home" ui-sref="home">Home</a> | <a href="#/about" ui-sref="about">About</a> | <a href="#/contacts?page=2" ui-sref="{page: 2}">Next page</a>
- * 
+ *
  * <ul>
  *     <li ng-repeat="contact in contacts">
  *         <a href="#/contacts/1" ui-sref="contacts.detail({ id: contact.id })">Joe</a>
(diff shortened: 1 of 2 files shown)


src/stateDirectives.js
```diff
@@ -279,7 +279,61 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
...
+/**
+ * @ngdoc directive
+ * @name ui.router.state.directive:ui-state
+ *
+ * @requires ui.router.state.uiSref
+ *
+ * @restrict A
+ *
+ * @description
+ * Much like ui-sref, but will accept named $scope propertys to evaluate for a state definition,
```
nateabele: s/propertys/properties/

src/stateDirectives.js
```diff
@@ -279,7 +279,61 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
...
+        angular.extend(args[2], {
+          uiSref:     buildUiSref(),
+          uiSrefOpts: attrs.uiStateOpts,
+        });
+
+        uiSrefDirective[0].link.apply(null, args);
+      }
+
+      scope.$watch(function () {
+        return scope.$eval(attrs.uiState) +
```
nateabele: You _should_ be able to convert this to:

```javascript
scope.$watch(function () {
  return [attrs.uiState, attrs.uiStateParams, attrs.uiStateOpts];
}, runStateRefLink, true);
```

nateabele:
Hey, thanks for pulling this together! I made a couple of comments, but barring those it looks good to me. @christopherthielen Any comments?

christopherthielen:
I like the effort to reuse the existing ui-sref code for ui-sref-active, (hopefully that works).  I'll have to play with this for a while before I can comment on the implementation.

src/stateDirectives.js
kasperlewau: I'll get on it asap, thanks for the catch :)

src/stateDirectives.js
kasperlewau: I like that, cutting away at the amount of $eval!

But - I do think we have to keep at least one $eval in there, otherwise changes wouldn't be picked up if the underlying data bound to the scope variable is modified - only so if the *actual* variable in the ui-state definition changes. 

So, my suggestion would be: 

```js
scope.$watch(function () {
  return scope.$eval([attrs.uiState, attrs.uiStateParams, attrs.uiStateOpts]);
}, runStateRefLink, true);
```

What do you think about that? @nateabele

kasperlewau:
Hmph. No sauce. evaluating the full array will always return undefined. 

I can think of a couple of ways to achieve the expected result and move away from concatenation: 

```js
  return [scope.$eval(attrs.uiState), scope.$eval(attrs.uiStateParams), scope.$eval(attrs.uiStateOpts)];

  return [scope[attrs.uiState], scope[attrs.uiStateParams], scope[attrs.uiStateOpts]];

  return [attrs.uiState, attrs.uiStateParams, attrs.uiStateOpts].map(function (prop) {
    return scope.$eval(prop);
  });  
```

thoughts?

nateabele:
> Hmph. No sauce. evaluating the full array will always return undefined.

Yeah, I just realized I was looking at it wrong... hmm...

> I can think of a couple of ways to achieve the expected result and move away from concatenation:

I'm not so much worried about the syntax as I am about the performance. Trying to think of how we can get it down to one `$watch()` / `$eval()`. Those are going to add up _really_ quickly. I wonder if `attrs.$observe()` would be any cheaper...

I'll test it out when I have time, unless you beat me to it.

kasperlewau:
I'll see if I can come up with some way of cutting it down to a single $eval, not sure what that would look like though. 

Regarding `attrs.$observe` vs `scope.$watch` - I do *believe* that the former is slightly faster, but I'm not 100% on that one. In any case I think the difference would be negligible, it's the amount of $eval's that are worrying.

kasperlewau:
Reverted the previous to 3x `$eval`, so as to get a passing build. Still haven't figured out a better way to evaluate the three values without three consecutive calls to `$eval` or `$parse` (albeit I haven't worked on this full time...).

kasperlewau:
Unfortunately I have *still* not been able to figure out a way to run a single $eval call for three separate expressions - without parsing possible objects into JSON and then concatenating the strings. 

I did clean up the code ever so slightly in the `$watch` fn expression from: 
```js
return [
  scope.$eval(attrs.uiState),
  scope.$eval(attrs.uiStateParams),
  scope.$eval(attrs.uiStateOpts)
];
```

Into: 

```js
return [
  attrs.uiState,
  attrs.uiStateParams,
  attrs.uiStateOpts
].map(scope.$eval.bind(scope));
```

--- 

Opted for `angular.bind` so as to get a passing spec on TravisCI (all passing locally with `fn.proto.bind` running `karma integrate`). 

*One could argue that it "helps" with ES5+IE8, but that is sort of moot seeing as how `.map` is being utilised.*

kasperlewau:
@nateabele @christopherthielen - What are your thoughts on the current state of things in this PR? Have you gotten a chance to play around with it?

---

Personally, I will *probably* never use `ui-state` (it was always just about building it for me), so I'm not in a rush. Though I have been receiving some questions as to the status of the PR and when a possible merge would occur - I'm sort of at a loss on what to respond with (barring the `ah, just +1 it friend`). 

I do share your concern in regards to multiple calls to `$eval`, and I have yet (read to test out the `$attrs.observe` approach. Even so I'm not crystal clear on how one would measure the possible difference in performance, any and all help in figuring that one out would be very appreciated.

nateabele:
Refactored this into #2187.

kasperlewau:
Closing this in favour of #2187.

PR #1934, opened 2015-05-04
feat($state): Inject templateProvider with resolved values

### LINKED ISSUE #330: Unable to inject templateProvider with resolved values
A view's `templateProvider` is unable to be injected with resolved values. The following will fail with an error like `Error: [$injector:unpr] Unknown provider: barProvider <- bar`.

```javascript
$stateProvider.state('foo', {
  resolve: {bar: function() { return 123; }},
  views: {
    foo: {
      templateProvider: function(bar) { /* ... */ },
    }
  },
})
```

I see in `resolveState` that `$view.load` only gets some locals to inject. Is there any technical reason why `templateProvider` couldn't be injected with all of the resolve values in the same way that `controller` is?

(This is on git master, and I remember seeing the same limitation a couple of months ago.)

webgago:
Fixes #330

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -1390,30 +1390,38 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
       })];
       if (inherited) promises.push(inherited);
 
-      // Resolve template and dependencies for all views.
-      forEach(state.views, function (view, name) {
-        var injectables = (view.resolve && view.resolve !== state.resolve ? view.resolve : {});
-        injectables.$template = [ function () {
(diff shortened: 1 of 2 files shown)


src/state.js
```diff
@@ -1385,33 +1385,41 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
...
+              result.$$controller = view.controller;
+            }
+            // Provide access to the state itself for internal use
+            result.$$state = state;
+            result.$$controllerAs = view.controllerAs;
+            dst[name] = result;
+          }));
+        });
+
+        return $q.all(viewsPromises);
```
webgago: should I return dst here?

webgago:
thank you!

Issue #1915, opened 2015-04-24
0.2.14 Caused a breaking change


OverZealous:
I (accidentally) just upgraded to the latest release, and suddenly my tests won't pass with this error:

```plain
 Error: Default value (false) for parameter 'error' is not an instance of Type (any)
          at $$getDefaultValue (./vendor/angular-ui-router/release/angular-ui-router.js:1628)
          at $value (./vendor/angular-ui-router/release/angular-ui-router.js:1643)
          at ./vendor/angular-ui-router/release/angular-ui-router.js:1686
          at forEach (./vendor/angular/angular.js:323)
          at ./vendor/angular-ui-router/release/angular-ui-router.js:1687
          at transitionTo (./vendor/angular-ui-router/release/angular-ui-router.js:3148)
          at go (./vendor/angular-ui-router/release/angular-ui-router.js:3069)
          at ./build/connect/js/app/routes/error/error.route.js:5:0 <- null:0:0
          at check (./vendor/angular-ui-router/release/angular-ui-router.js:2008)
          at update (./vendor/angular-ui-router/release/angular-ui-router.js:2020)
          at ./vendor/angular/angular.js:14712
          at ./src/connect/app/routes/routes.spec.js:30
```

I don't mind fixing the issue (it worked without error on 0.2.13), but this is a breaking change on a what should be a patch release.

The faulty code in question looks like this:
```js
$stateProvider.state('error', {
    params: {
        error: false,
        fromState: {},
        fromParams: {},
        toState: {},
        toParams: {},
    },
   //...
```

-----

Actually, on reading through the docs again, this seems like it should be an acceptable format, so this might actually be a legitimate bug.

nateabele:
Yeah, seems like it might be a bug. I'll investigate shortly.

eddiemonge:
```
.config ( $stateProvider ) ->
  $stateProvider

  .state 'packages',
    abstract: true
    parent: 'layout'
    template: '<ui-view />'
    url: '/packages/'

  .state 'packages.received',
    url: 'received?page&per_page&archived&q'
    reloadOnSearch: false
    params: 
      page:
        squash: true
        value: 1
      per_page:
        squash: true
        value: 10
      archived:
        squash: true
        value: false
      q: squash: true
    templateUrl: '/modules/packages/index.html'
    controller: 'PackagesIndexController'
     
```

*sidenote* squash never seemed to squash the query string part even when it matched. im guessing its because of type as well

[...]

PR #1944, opened 2015-05-07
fix($state) The `is` check on type `any` will always return true.

ckniffen:
Fixes breaking change in 0.2.14 where parameters no longer were allowed to have default values of `0`, `false`, `''`, or `undefined`.

Fixes: #1915

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -623,7 +623,6 @@ function $UrlMatcherFactory() {
     any: { // does not encode/decode
       encode: angular.identity,
       decode: angular.identity,
-      is: angular.identity,
       equals: angular.equals,
       pattern: /.*/
     }
--- a/test/stateSpec.js
+++ b/test/stateSpec.js
@@ -24,6 +24,7 @@ describe('state', function () {
       DD = { parent: D, params: { x: null, y: null, z: null } },
       DDDD = { parent: D, controller: function() {}, template: "hey"},
       E = { params: { i: {} } },
+      F = { params: { a: '', b: false, c: 0, d: undefined, e: -1 }},
       H = { data: {propA: 'propA', propB: 'propB'} },
       HH = { parent: H },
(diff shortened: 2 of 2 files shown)


src/urlMatcherFactory.js
```diff
@@ -623,7 +623,7 @@ function $UrlMatcherFactory() {
     any: { // does not encode/decode
       encode: angular.identity,
       decode: angular.identity,
-      is: angular.identity,
+      is: function(){ return true; },
```
eddiemonge: a function that always returns true?

test/stateSpec.js
```diff
@@ -24,6 +24,7 @@ describe('state', function () {
       DD = { parent: D, params: { x: null, y: null, z: null } },
       DDDD = { parent: D, controller: function() {}, template: "hey"},
       E = { params: { i: {} } },
+      F = { params: { a: '', b: false, c: 0, d: undefined }},
```
eddiemonge: `-1` as well for fun?

ckniffen: @eddiemonge I can definitely add -1.  Also I added the test here because I could not find a place where the is method on types were tested.  Is there a better place for this test.  And maybe instead of testing by navigating to a state with default params all of the `is` methods should be tested separately.

src/urlMatcherFactory.js
ckniffen: @eddiemonge The reason for this is that any was always returning whatever you passed in.

`any.is(false)` would always return `false`.  Since everything is type `any` it should always return true.

nateabele:
Yeah, looks good. Throw in `-1` for the heck of it and squash your commits. Thanks again for the patch.

Strictly speaking, the test should probably go against the spec for `$urlMatcherFactory`, but this is totally fine for now.

eddiemonge:
You changed the commit message style so it doesnt follow angular guidelines anymore

ckniffen:
@eddiemonge Sorry about that.  I have updated the message.

PR #1947, opened 2015-05-10
feat(isState): add support for `params` and `options` for `isState` filter

julienevano:
Add support for `params` and `options` optional arguments for `isState`
filter in order to have the same functionality as `$state.is(stateName,
params, options)`.
e.g.:
- `'a.c' | isState:{params: {x: 'foo', y: 'bar'}}` in the case of
  `$stateProvider.state('a.c', { url: '/:x?y' })`
- `'.c' | isState:{params: {x: 'foo', y: 'bar'}, options: options}` with `$scope.options =
  { relative: $state.get('a') }` in the case of
  `$stateProvider.state('a.c', { url: '/:x?y' })`

### DIFF
--- a/test/stateFiltersSpec.js
+++ b/test/stateFiltersSpec.js
@@ -3,7 +3,10 @@ describe('isState filter', function() {
   beforeEach(module(function($stateProvider) {
     $stateProvider
       .state('a', { url: '/' })
-      .state('a.b', { url: '/b' });
+      .state('a.b', { url: '/b' })
+      .state('a.c', { url: '/:x?y' })
+      .state('d', { url: '/' })
+      .state('d.c', { url: '/c' });
   }));
 
   it('should return true if the current state exactly matches the input state', inject(function($parse, $state, $q, $rootScope) {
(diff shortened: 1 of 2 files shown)


test/stateFiltersSpec.js
```diff
@@ -17,6 +20,34 @@ describe('isState filter', function() {
...
+  }));
+
+  it('should work for relative states', inject(function ($parse, $q, $state, $rootScope) {
+    var scope = $rootScope.$new();
+    scope.options = { relative: $state.get('a') };
+
+    $state.go('a.c');
+    $q.flush();
+
+    expect($parse('".c" | isState:undefined:options')(scope)).toBe(true);
```
nateabele: Having to do `:undefined` seems a little tedious. Why not change `isFilter()` to do `params || undefined` or something?

julienevano: `isState` filter has been updated to have only one optional parameter as an object, e.g. `".c" | isState:{params:{x: "foo", y: "bar"}, options: options}`

nateabele:
That seems a lot more verbose, and doesn't really match up to the underlying API. Is there not an easy way to do 'skipped' parameters in filters?

julienevano:
I was thinking to test the `arguments` passed in `isFilter` but `params` and `options` are both an Object so I cannot differentiate which one is passed in the case when we want to pass only one argument (e.g. `'stateName' | isState:params` vs `'stateName' | isState:options`). If you know a way to skip parameters in filters, could you give me a clue? Thank you.

Issue #582, opened 2013-11-14
$state.reload() should reinitialize the controller


ghengeveld:
As discussed in #76, $state.reload() does reload the resolves, but doesn't reinitialize the controller with the newly resolved values. Since the whole point of reloading the resolves is to refresh the scope with the new data, reinitializing the controller is a requirement for $state.reload() to be of any use.

Demo: http://plnkr.co/edit/GNr1bN?p=preview

softBarbarian:
Actually I soved differently: basically I fetch data $ionView.loaded lifecycle event

hpawe01:
If you are using the current **ionicframework** (ionic: v1.0.0-beta.14, angularjs: v1.3.6, angular-ui-router: v0.2.13), the problem with the not-reloading-controller could be caused by the [new **caching**-system of ionic](http://ionicframework.com/docs/api/directive/ionNavView/):
>Note that because we are caching these views, we aren’t destroying scopes. Instead, scopes are being disconnected from the watch cycle. Because scopes are not being destroyed and recreated, **controllers are not loading again** on a subsequent viewing.

There are several ways to disable caching. To disable it only for a single state, just add `cache: false` to the state definition.

This fixed the problem for me (after hours of reading, trying, frustration).

For all others not using ionicframework and still facing this problem: good luck!

gregor-srdic:
Thank you very much for this post hpawe01, I lost quite a few hours due to this new feature and cache: false worked perfectly!

malixsys:
@hpawe01 Indeed! +1

    $stateProvider.
      state('home', {
        url: '/',
        cache: false,

oozzal:
Thanks @hpawe01. Certainly saved me few hours. :clock10:

munkychop:
+1 for an official fix please.

The proposed milestone of this issue (at the time of writing) is 0.2.12 and the current release version of ui-router is 0.2.13. Is the issue likely to be fixed anytime soon? If so, when roughly?

ui-router is amazing! But this issue has been open for over a year...

christopherthielen:
This issue was closed 2 months ago

munkychop:
Oh, my apologies, I assumed it needed re-opening due to the posts from a few days ago.

Thanks for the fix :)

ehlerskon:
thanks @hpawe01 - you saved my day!

pholly:
[...]

PR #1962, opened 2015-05-18
fix($state): reloadOnSearch should not affect non-search param changes.

### LINKED ISSUE #1079: Problem with reloadOnSearch: false - change in stateParams doesn't reload
Here is my issue:

I navigate through a list of links, that display different images depending on the date in the search parameters (?from&to). 

Those search parameters should not trigger a state reload. Finally, here is my state declaration:
```javascript
state('links/:linkId', {
  url: "/links/linkId?from&to",
  reloadOnSearch: false,
  templateUrl: ...
  Controller: ...
});
```

But then a huge problem shows up! When I try to navigate through my different links, reload doesn't happen anymore, so that I cannot pass from one link to another.

After some digging in the code, I found out that when `reloadOnSearch` equals to false, triggers are always blocked because there's absolutely no checking of whether the changing parameters are search parameters or path parameters.

So actually, `reloadOnSearch` doesn't work as intended and should be renamed `reloadOnStateParamsChange`.

I fixed the issue by checking if the state parameters that changed are search parameters or not, which seems to work fine.

Please tell me if I'm mistaking somewhere or if it is relevant, and also if you would be interested in a pull request or a plunkr. 

Thanks in advance!

metamatt:
The handling of `reloadOnSearch: false` caused ui-router to avoid
reloading the state in more cases than it should. The designed and
documented behavior of this flag is to avoid state reload when the
URL search string changes. It was also avoiding state reload when
the URL path (or any non-search parameters to the state) changed,
and even when state reload was explicitly requested.

This change

- flips the name of shouldTriggerReload (and the accompanying guard
  boolean, skipTriggerReloadCheck) to match the direction of the
  logic: shouldSkipReload and allowSkipReloadCheck

- teaches shouldSkipReload to look at the types of the differing
  parameters, and only skip the reload if the only parameters that
  differ were search parameters

- pulls the test for options.reload to the front of the complex
  boolean expression. (I think calling $state.reload() explicitly
  should reload a state even if it has reloadOnSearch:false. But I
  don't understand exactly why the test for this was where it was, and
  maybe there's a good reason I'm missing. Also, if the behavior I
  favor is deemed correct, we could also achieve that by setting
  allowSkipReloadCheck=false for all truthy values of options.reload,
  namely, if options.reload===true, and then shouldSkipReload wouldn't
  even need to look at options. I left a comment about this in the
  source too and would appreciate feedback.)

Fixes #1079. Helps with one of the cases broken in #582.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -954,7 +954,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
      *    have not changed, aka a reload of the same state. It differs from reloadOnSearch because you'd
      *    use this when you want to force a reload when *everything* is the same, including search params.
      *    if String, then will reload the state with the name given in reload, and any children.
-     *    if Object, then a stateObj is expected, will reload the state found in stateObj, and any chhildren.
+     *    if Object, then a stateObj is expected, will reload the state found in stateObj, and any children.
      *
      * @returns {promise} A promise representing the state of the new transition. See
      * {@link ui.router.state.$state#methods_go $state.go}.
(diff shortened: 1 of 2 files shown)


metamatt:
Review notes: What's here works for my purposes and passes the test suite including a new test for the #1079 case it fixes. It may not be the optimal way of spelling out these concepts for your codebase and I'm happy to adjust this with feedback and get it right.

christopherthielen:
I modified and manually merged this in

PR #1960, opened 2015-05-18
refactor(commonjs): Added support for commonjs

aaronroberson:
Added `ui.router` to module exports so that after installing via npm, the package can be required in the angular module dependency list such as:

```angular.module('myApp', [require('angular-ui-router')]);```

This is helpful for Browserify and Webpack.

PR #1970, opened 2015-05-20
Revert "refactor(commonjs): Added support for commonjs"

nateabele:
Reverts angular-ui/ui-router#1960

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,5 @@
     "grunt-conventional-changelog": "~1.1.0",
     "grunt-ngdocs": "~0.2.5"
   },
-  "main": "index.js"
+  "main": "release/angular-ui-router.js"
 }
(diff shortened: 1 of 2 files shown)


nateabele:
@eddiemonge Good catch, not sure how I didn't realize that. Here you go.

package.json
```diff
@@ -62,5 +62,5 @@
     "grunt-conventional-changelog": "~1.1.0",
     "grunt-ngdocs": "~0.2.5"
   },
-  "main": "index.js"
+  "main": "release/angular-ui-router.js"
```
eddiemonge: should it be `"release/angular-ui-router.js"` or `"./release/angular-ui-router.js"`? Does it make a difference? Bower has it one way and component.json and package.json the other.

nateabele:
> Does it make a difference?

Probably not.

> Bower has it one way and component.json and package.json the other.

I guess we can patch them to be consistent, but I don't think it's that big a deal.

PR #1974, opened 2015-05-20
Update common.js

AlexJuarez:
Fixes an error that can occur if parents[i] is undefined.

### DIFF
--- a/release/angular-ui-router.js
+++ b/release/angular-ui-router.js
@@ -109,7 +109,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
+    if (!parents[i] || !parents[i].params) continue;
     parentParams = objectKeys(parents[i].params);
     if (!parentParams.length) continue;
 
--- a/src/common.js
+++ b/src/common.js
@@ -96,7 +96,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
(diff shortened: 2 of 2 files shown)


release/angular-ui-router.js
```diff
@@ -109,7 +109,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
+    if (!parents[i] || !parents[i].params) continue;
```
eddiemonge: [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) might be a better check. Also, the correct place to do this fix is at https://github.com/angular-ui/ui-router/blob/master/src/common.js#L99

src/common.js
```diff
@@ -96,7 +96,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
+    if (!parents[i] || !parents[i].params) continue;
```
eddiemonge: [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) might be a better check

release/angular-ui-router.js
```diff
@@ -4367,4 +4367,4 @@ function $IncludedByStateFilter($state) {
 angular.module('ui.router.state')
   .filter('isState', $IsStateFilter)
   .filter('includedByState', $IncludedByStateFilter);
-})(window, window.angular);
\ No newline at end of file
+})(window, window.angular);
```
eddiemonge: don't include this change

eddiemonge:
See the notes at https://radify.github.io/PR.js/#/angular-ui/ui-router/pulls/1974

PR #1975, opened 2015-05-21
fix(common.js): Error if parents[i] is undefined

AlexJuarez:
hasOwnProperty still would throw an error if parents[i] was undefined.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -96,7 +96,7 @@ function inheritParams(currentParams, newParams, $current, $to) {
   var parents = ancestors($current, $to), parentParams, inherited = {}, inheritList = [];
 
   for (var i in parents) {
-    if (!parents[i].params) continue;
+    if (!parents[i] || !parents[i].params) continue;
     parentParams = objectKeys(parents[i].params);
     if (!parentParams.length) continue;
 


eddiemonge:
What case would `parents[i]` be undefined since `var i in parents`?

AlexJuarez:
The parents array can be populated with undefined, right now I'm working on a app that im switching from ember router to angular ui router, and the ember router pushes undefined into the array.

eddiemonge:
That sounds like a bug that needs to be fixed there but having the check here isn't bad. The check itself could be improved though. I think it should be something like `parents[i] && !parents[i].params`.

AlexJuarez:
Changing this line allows both routers to co-exist

AlexJuarez:
What about typeof parents[i] === 'undefined' || !parents[i].params

AlexJuarez:
I would still like to get this fix in, even though its low priority.

christopherthielen:
I'm ok with merging this in if we get a unit test, but ...

>Changing this line allows both routers to co-exist

...is obviously not a goal of this project

christopherthielen:
Whoops, looks like #2478 was an exact dupe of this one.

PR #1986, opened 2015-05-23
feat(uiSref): add ui-sref-on attribute

### LINKED ISSUE #1863: RFC: Add a ui-sref-resolve directive similar to ui-sref-active(-eq)

wms:
Fixes #1863, supersedes #1877

### DIFF
--- a/test/stateDirectivesSpec.js
+++ b/test/stateDirectivesSpec.js
@@ -389,6 +389,42 @@ describe('uiStateRef', function() {
       expect(transitionOptions.notify).toBeUndefined();
     }));
   });
+
+  describe('transition expression', function() {
+    beforeEach(inject(function($rootScope, $compile) {
+      el = angular.element('<a ui-sref="contacts.item.detail({id: 5})" ui-sref-on="loading = $transition">Details</a>');
+      scope = $rootScope;
+      scope.loading = null;
+
+      $compile(el)(scope);
+      scope.$digest();
+    }));
+
+    it('applies an expression when a transition begins', inject(function($rootScope, $timeout, $state) {
+      if (angular.version.minor < 2) return;
+
+      var newState;
+
+      expect(scope.loading).toBe(null);
+      triggerClick(el);
+      $timeout.flush();
+
+      expect(scope.loading.then).toEqual(jasmine.any(Function));
+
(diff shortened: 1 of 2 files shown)


test/stateDirectivesSpec.js
```diff
@@ -389,6 +389,42 @@ describe('uiStateRef', function() {
       expect(transitionOptions.notify).toBeUndefined();
     }));
   });
+
+  describe('transition expression', function() {
+    beforeEach(inject(function($rootScope, $compile) {
+      el = angular.element('<a ui-sref="contacts.item.detail({id: 5})" ui-sref-on="loading = $transition">Details</a>');
+      scope = $rootScope;
+      scope.loading = null;
```
nateabele: @warrenseymour I thought of a way to fix the tests for the 1.0.x thing until we drop support for it:

```javascript
var _transition = null;
el = angular.element('<a ui-sref="contacts.item.detail({id: 5})" ui-sref-on="load($transition)">Details</a>');
scope.load = function(transition) {
  _transition = transition;
}
```

Then s/`scope.loading`/`_transition`/.

TimonVS:
When will this be merged in?

PR #1991, opened 2015-05-27
fix($state): Fixed bug within resolveState() method

chris1705:
Change the second parameter of $resolve.resolve in line 3788 from `dst.globals` to `locals` to make resolving work as expected.

Additional information: Commit afa20f22 changed the resolveState() method. It broke resolving promises, so they were not injected to the controller. I got the following error:

```
Error: [$injector:unpr] Unknown provider: assetProvider <- asset <- VisualizationOverviewCtrl
[...]
```


nateabele:
Hmm... I'm pretty sure we have unit tests that cover this. Can you write a unit test (or at least put together a plunkr) that demonstrates this?

PR #2002, opened 2015-05-29
ui-required attribute

### LINKED ISSUE #2001: Feature Request: conditional ui-views based on url parameters
As of now it is not posible in a useful way to define ui-views inside ng-repeat. Imagine a list view where the list items expand to a detail view. In this case it would be nice to have conditional ui-views something like this:

```javascript
<div ng-repeat="item in list">
    <div ui-view="detail" condition="{slug:item.id}"></div>
</div>
```

jimmypoms:
resolve #2001 by adding a ui-required attribute where it is possible to specify url parameters to be matched against the route.

### DIFF
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -209,7 +209,7 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate)
 
         function updateView(firstTime) {
           var newScope,
-              name            = getUiViewName(scope, attrs, $element, $interpolate),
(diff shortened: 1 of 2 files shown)


christopherthielen:
Have you considered something like
```
<div ng-repeat="thing in allthings">
  <div ui-view ng-if="$stateParams.id == thing.id">
```

jimmypoms:
Didn't think that this would work. Will give it a shot :)

jimmypoms:
It does work with the ng-if, although it feels less fancy ;). Anyhow closing the PR

PR #2003, opened 2015-05-29
fix(ui-view): change $viewContentLoading to pair with $viewContentLoaded

### LINKED ISSUE #685: event $viewContentLoading doesn't seem to work
It seems I'm unable to get the `$viewContentLoading` fired.

The doc says it's fired by the `$rootScope`, while in the example listen to the `$scope`

```
$scope.$on('$viewContentLoading', 
function(event, viewConfig){ 
    // Access to all the view config properties.
    // and one special property 'targetView'
    // viewConfig.targetView 
});
```

I've tried both, but I couldn't catch any event

metamatt:
The old $viewContentLoaded event wasn't needed for the one internal
client (ui-view) because ui-view always ignores that event and acts on
a $stateChangeSuccess event that follows right behind anyway.

And it wasn't as useful to external clients as it could be because
it wasn't delivered on every view update -- it was delivered only
on state transitions that activate a state defining a view, and
didn't deal with inheritance.

Also, neither $viewContentLoading nor $viewContentLoaded events
contained the name of the view loading or loaded.

This change makes $viewContentLoading and $viewContentLoaded be
emitted always in pairs, before/after updateView does the work of
actually loading the view, and both events include the name of the
view being loaded.

This is a breaking change for users of the old $viewContentLoading
event that relied on it being broadcast from the root scope, the
precise time it was broadcast, the fact that it wasn't always sent
when a view is loaded even if $viewContentLoaded will be sent, or the
"options" parameter that was emitted with it. If people rely on any of
that behavior and we can't break it, then we can restore the old
behavior but I think there needs to be some other event which is
reliably paired with $viewContentLoaded and contains the view name,
and I can't think of a better name than $viewContentLoading for that
event.

Closes #685.

### DIFF
--- a/src/view.js
+++ b/src/view.js
@@ -36,32 +36,6 @@ function $ViewProvider() {
         if (options.view) {
           result = $templateFactory.fromConfig(options.view, options.params, options.locals);
         }
-        if (result && options.notify) {
-        /**
-         * @ngdoc event
-         * @name ui.router.state.$state#$viewContentLoading
-         * @eventOf ui.router.state.$view
-         * @eventType broadcast on root scope
-         * @description
-         *
-         * Fired once the view **begins loading**, *before* the DOM is rendered.
-         *
-         * @param {Object} event Event object.
-         * @param {Object} viewConfig The view config properties (template, controller, etc).
-         *
-         * @example
-         *
-         * <pre>
-         * $scope.$on('$viewContentLoading',
-         * function(event, viewConfig){
-         *     // Access to all the view config properties.
-         *     // and one special property 'targetView'
-         *     // viewConfig.targetView
-         * });
-         * </pre>
-         */
-          $rootScope.$broadcast('$viewContentLoading', options);
-        }
         return result;
       }
     };
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -180,9 +180,6 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate)
         scope.$on('$stateChangeSuccess', function() {
           updateView(false);
         });
-        scope.$on('$viewContentLoading', function() {
-          updateView(false);
-        });
 
         updateView(true);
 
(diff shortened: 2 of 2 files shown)


kevinob11:
I want this.

christopherthielen:
> The old $viewContentLoaded event wasn't needed for the one internal client (ui-view) because ui-view always ignores that event and acts on a $stateChangeSuccess event that follows right behind anyway.

@metamatt thank you, I think your analysis is correct.  I think this change would also address some issues with `notify: false`, as some other users mentioned.

> This is a breaking change for users of the old $viewContentLoading event that relied on it being broadcast from the root scope, the precise time it was broadcast, the fact that it wasn't always sent
when a view is loaded even if $viewContentLoaded will be sent, or the "options" parameter that was emitted with it. 

meh.  I can't think of a scenario where you'd care about those things, except perhaps the options or $rootScope. I'm sure the one guy who relies on this will chime in to let us know eventually.

@nateabele  if we accept this PR, let's bump our next release to 0.3.0 just to keep the semver folks off our backs.  Also, we'd need to merge the relevant bits into `feature-1.0` branch.

christopherthielen:
@kevinob11 @metamatt what do you use $viewContentLoading/Loaded for?

kevinob11:
I don't use it at all. The issue I was running into was views reloading twice after a notify:false view change.

metamatt:
My use case:

> my test harness wants to know when views are instantiating so that if a controller throws an exception, it knows which view to blame. So I really want a reliable "loading" signal.

(this and some additional thoughts are in https://github.com/angular-ui/ui-router/issues/685#issuecomment-106621834)

metamatt:
And re

> meh. I can't think of a scenario where you'd care about those things

I 100% agree. I was just being pedantic about describing the change in behavior to give anyone who *would* care a chance to notice and speak up. I would expect that anyone who is using these events would consider this change an improvement.

nateabele:
@christopherthielen Just merged this. I don't remember what my objection was before, but it makes sense to me now.

btm1:
when will the next release happen? I want this :+1:

rcollette:
+1  Need a release.  The lack of consistent event pairs is a problem for me.

chence:
+1 need a release

genu:
+1 for release

Barnzy:
+1 for release :+1:

EvilJohnny:
+1 for release! :smiley:

Rebing:
+1 for release

tchiloh:
when will the next release happen?

Xaiid:
+1 release please :(

skomma:
+1 for release!

kouki-o:
+1 please release.

davidouanounou:
+1 for release !

Issue #1973, opened 2015-05-20
Slashes are double encoded in hash query params


natefaubion:
I have a link that's something like
```html
<a ui-sref="myState({ path: '/foo' })">Go</a>
```

Which results in the hash-url `#/my-state?path=%252Ffoo`. The slash being encoded to `%2F` and the percent being encoded to `%25`. In `0.2.13` this double-encoding seemed to work fine, as the new state picks it up correctly (as `/foo`), but in `0.2.14` and `0.2.15` the new state will only get it decoded once, resulting in `%2Ffoo`.

natefaubion:
I've actually found that ui-router is transitioning twice on this link. If I don't have a slash in the query param, it transitions once. But once I add the slash it does two transitions, one immediately after the other. The first transition get the correct param, and the second is double encoded.

natefaubion:
I've also found that if you provide a direct url like `#/my-state?path=/foo` (not through a ui-sref link), the router automatically navigates to `#/my-state?path=%2Ffoo` (not double-encoded) which works, but inserts a new history item, which breaks the back button (when you hit back it just keeps redirecting you over and over again).

ckniffen:
This is related to #1645.  Before 0.2.14 the issue was more cosmetic but with the addition of the `$normalize` method on types the decode function will no longer be called for params of type `string`.

I have created a failing test for it.
```
it("should encode and decode slashes in parameter values", function () {
    var matcher = new UrlMatcher('/:foo');
    expect(matcher.format({ foo: "/" })).toBe('/%252F');
    expect(matcher.format({ foo: "//" })).toBe('/%252F%252F');
    expect(matcher.exec('/%2F').foo).toBe('/');
    expect(matcher.exec('/%2F%2F').foo).toBe('//');
 });
```
Previously the test did have assertions for decoding.

## Solution 1
Have the `string` type override `$normalize` to always call decode

```
string: {
      encode: valToString,
      decode: valFromString,
      $normalize: function(val){ return this.decode(val); },
      is: function(val) { return val == null || !isDefined(val) || typeof val === "string"; },
      pattern: /[^/]*/
}
```

## Solution 2

Update is method to `is` method to check for `\` encoding.

piupiupiu:
Unfortunately, it seems it is not fixed.

My configuration is the following:
```
$stateProvider
        .state('root',{
            url: '/{lang:(?:[A-Za-z]{2}/)?}',
            abstract: true,
            template: '<ui-view />',
[...]

Issue #1645, opened 2014-12-18
Slash encoding


theatrain:
I am still getting the problem when trying to pass dates.

    ui-sref="jobs-list({quote_due_from:'18/12/2014',quote_due_to:'18/12/2014'})"

The following URL is generated:

    http://{mydomain}/#/jobs?quote_due_from=18%252F12%252F2014&quote_due_to=18%252F12%252F2014

This is my state provider code:

    module.config(function($stateProvider) {
        $stateProvider
            .state('jobs-list', {
                url: '/jobs?quote_due_from&quote_due_to&quote_followup_from&quote_followup_to&start_date_from&start_date_to&quote_status_id',
                controller: 'JobsListCtrl',
                templateUrl: 'jobs/jobs.list.html'
            });
    });

I am using the latest version of ui.router

    "name": "angular-ui-router",
    "version": "0.2.13",

kcjmowright:
I got around this issue by explicitly configuring the parameter types to 'any'.

```
module.config(function($stateProvider) {
    $stateProvider
        .state('jobs-list', {
            url: '/jobs?{quote_due_from:any}&{quote_due_to:any}&{quote_followup_from:any}&{quote_followup_to:any}&{start_date_from:any}&{start_date_to:any}&quote_status_id',
            controller: 'JobsListCtrl',
            templateUrl: 'jobs/jobs.list.html'
        });
});
```

tweinerpq:
Hi,
I'm still experiencing double encoding when the string I'm sending in is part of the URL, like so:
example.com/id

I'm sending in this:
<a ui-sref="search.doc({ id: 'CBVPPYMD10.1017/S003329171400227X' })" ...

and on click I'm seeing on my browser

url.com/editor/CBVPPYMD10.1017%252FS003329171400227X

I'm also seeing this same doubly encoded string as the href attribute generated by ui-router on Inspect Element.

My real doc id is CBVPPYMD10.1017/S003329171400227X

I just updated to @version v0.2.15 and I have latest angular (1.4)

Thanks.

tweinerpq:
Oh, I see that this is being addressed here?
https://github.com/angular-ui/ui-router/pull/2071

Is this supposed to be fixed?

chliebel:
[...]

PR #2005, opened 2015-05-30
fix(urlMatcherFactory): Slashes are now properly decoded in url params

ckniffen:
Fixes: #1973, #1645


piotrd:
:+1: tested with a large app, works great. Can anyone merge it, please?

ckniffen:
@piotrd have you checked #2071? I think it is a better implementation.

piotrd:
@ckniffen: you're right. Thanks for pointing this one out, I somehow didn't notice that PR. I tested it now and unfortunately it fails for `val` being null. I made a comment in the PR about it, along with proposed improvement (https://github.com/angular-ui/ui-router/pull/2071/files#r33758421).

ckniffen:
Closing in favor of #2071.

@piotrd I get that too.  I could not find a way around it though.

Issue #1997, opened 2015-05-28
Using ui-sref-active[-eq] with ng-repeat causes a race condition


joshuahiggins:
As the title states, when using `ui-sref-active` or `ui-sref-active-eq` on a dynamically generated element, such as one created via `ng-repeat`, the element is not always ready for classes to be applied when `$stateChangeSuccess` is fired. This causes the classes to not apply on the initial state load, but they are applied on subsequent state changes.

Sample markup:

```html
<ul>
  <li ng-repeat="nav in vm.navItems" ui-sref-active-eq="current" class="{{nav.classes}}">
    <a ui-sref="{{nav.state}}">
      <span ng-bind="nav.text"></span>
    </a>
  </li>
</ul>
```

Although UI Router matches the state properly, the `update()` fails to apply the classes as `$element` is not yet available on the DOM.

In the above code, removing the `{{nav.classes}}` resolves the issue, which leads me to believe it's simply a race condition between the DOM and UI Router. Wrapping the `$element.addClass(activeClass)` in a `$timeout` resolves the issue, though it causes numerous unit tests to fail.

joshuahiggins:
This issue blocks pull request #1999.

eddiemonge:
Sidenote: `class="{{nav.classes}}"` is kind of bad on its own. It should really be `ng-class="nav.classes"`

joshuahiggins:
Good point. Still, I was just illustrating that in some conditions the the DOM is ready while in some it's not. In my own code, there is no extra directives on the `ng-repeat` elements but the classes still aren't applied on initial load due to the classes being applied before the digest cycle has finished.

tiagoroldao:
I just got to this as we needed this very functionality and were running 0.2.13

While this solves part of the issue, it does not work for truly dynamic data - ui-sref elements will push to ui-sref-active[-eq], but do nothing when $destroy-ed

christopherthielen:
Plunker: http://plnkr.co/edit/TFmqKJln3cXKkAxS5PrW?p=preview

christopherthielen:
No longer using `$timeout`.  Now fixed by using `$evalAsync`
http://plnkr.co/edit/TFmqKJln3cXKkAxS5PrW?p=preview

PR #2012, opened 2015-06-04
feat(uiSrefActive): allow active & active-eq on same element

joshuahiggins:
This pull request will allow the usage of both `ui-sref-active` and `ui-sref-active-eq` on the same element. 

One use case includes menus with expanding states where a parent element can be both `active` and `open`, such as in this example:

```html
<ul>
  <li ui-sref-active="open" ui-sref-active-eq="active">
    <a ui-sref="home"></a>
    <ul>
      <li ui-sref-active-eq="active">
        <a ui-sref="home.view1"></a>
      </li>
      <li ui-sref-active-eq="active">
        <a ui-sref="home.view2"></a>
      </li>
    </ul>
  </li>
  <li ui-sref-active="open" ui-sref-active-eq="active">
    <a ui-sref="account"></a>
    <ul>
      <li ui-sref-active-eq="active">
        <a ui-sref="account.view1"></a>
      </li>
      <li ui-sref-active-eq="active">
        <a ui-sref="account.view2"></a>
      </li>
      <li ui-sref-active-eq="active">
        <a ui-sref="account.view3"></a>
      </li>
    </ul>
  </li>
</ul>
```

When the account child views are active, the parent view should have a class of "open". However, when the parent is active, then the parent element should have both "open" and "active" classes.

This pull request exposes and squashes a race condition bug that's present today, #1997.

I apologize that this is a duplicate of a pull request I previously closed. I trashed that request due to failing tests before learning the quirks of `$timeout.flush()` in Karma. After digging around, I noticed that you had a comment explaining a workaround. Reopened after getting all the tests sorted out and passing.

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -228,13 +228,14 @@ $StateRefActiveDirective.$inject = ['$state', '$stateParams', '$interpolate'];
 function $StateRefActiveDirective($state, $stateParams, $interpolate) {
   return  {
     restrict: "A",
-    controller: ['$scope', '$element', '$attrs', function ($scope, $element, $attrs) {
-      var states = [], activeClass;
+    controller: ['$scope', '$element', '$attrs', '$timeout', function ($scope, $element, $attrs, $timeout) {
+      var states = [], activeClass, activeEqClass;
 
       // There probably isn't much point in $observing this
       // uiSrefActive and uiSrefActiveEq share the same directive object with some
       // slight difference in logic routing
-      activeClass = $interpolate($attrs.uiSrefActiveEq || $attrs.uiSrefActive || '', false)($scope);
+      activeClass = $interpolate($attrs.uiSrefActive || '', false)($scope);
+      activeEqClass = $interpolate($attrs.uiSrefActiveEq || '', false)($scope);
 
       // Allow uiSref to communicate with uiSrefActive[Equals]
       this.$$addStateInfo = function (newState, newParams) {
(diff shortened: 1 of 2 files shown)


nateabele:
Seems legit.

>  apologize that this is a duplicate of a pull request I previously closed. I trashed that request due to failing tests before learning the quirks of $timeout.flush() in Karma. After digging around, I noticed that you had a comment explaining a workaround. Reopened after getting all the tests sorted out and passing.

So, all the extra calls to `timeoutFlush()` are because things fail otherwise?

joshuahiggins:
> So, all the extra calls to timeoutFlush() are because things fail otherwise?

The calls to `$timeout.flush()` are necessary as a `$timeout` was added to the directive to ensure that the element has finished rendering to the DOM before attempting to add a class to it. This is an issue that is currently present in the code and can be seen when using `ui-sref-active` in conjunction with `ng-repeat`, as well as other use cases. It just happens that this feature request highlighted that issue and caused the issue to happen more regularly.

That being said, running  `$q.flush()` -- which is just `$rootScope.digest()` -- after `$timeout.flush()` should be redundant as `$timeout` will trigger an `$apply()` which triggers `$rootScope.digest()` anyway. By that logic, I replaced `$q.flush()` with `timeoutFlush()` in all the `ui-sref-active[-eq]` tests and it passed all `karma:unit` tests, but failed some `karma:past` and `karma:unstable` failures. 

I am honestly not familiar enough with Angular pre 1.2.x versions to understand why  `$timeout.flush()` wouldn't satisfy the digest needs of these tests. On that note, I left `$q.flush()` in place and just added the `timeoutFlush()` method after on all tests relating to `ui-sref-active[-eq]` and all prior versions of Angular just..... sorted themselves out. All tests pass with flying colors, despite the redundancy of the flushes. Forgive my naivety on the topic!

nateabele:
@joshuahiggins No, it's fine, I was just checking. Thanks for the detailed explanation.

@christopherthielen Looks good to me. You? I don't think it should interfere with any future plans.

christopherthielen:
:+1:

joshuahiggins:
Any thoughts on when the next minor version release might be?

PR #2030, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "angular-ui-router",
-  "version": "0.2.15",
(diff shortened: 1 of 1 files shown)


eddiemonge:
Even though its temporarily not used, it could be used in other places. Also, until its dropped from the spec completely, leaving it doesn't hurt much

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #2058, opened 2015-06-22
feat(improved commonjs and es6 ready) Update library to export main module through CommonJS

zeddever:
The convention for exporting data with angular modules is to export the main module. This this even more important as ES6 is being adopted and the convention is to import modules which use a default export to expose the main module, then referring to the module's name using the 'name' property from the angular module API.

Since only the name of this module was exported, this would prevent the use of the module in the conventional ES6 way and force you to hardcode the dependency name.

This fix allows you to do:

```
import ModuleRouter from 'angular-ui-router';

let MyModule = angular.module('MyModule', [ModuleRouter.name]);

...

exports default MyModule;

```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -21,12 +21,12 @@ module.exports = function (grunt) {
     concat: {
       options: {
         banner: '<%= meta.banner %>\n\n'+
+                '(function (window, angular, undefined) {\n',
+        footer: '})(window, window.angular);'+ 
                 '/* commonjs package manager support (eg componentjs) */\n'+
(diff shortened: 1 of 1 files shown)


christopherthielen:
I believe it's more common for angular modules to default export the angular module name, is it not?   So a user can use the imported variable in the dependencies list:
```
import uiRouter from "uiRouter";
import uiBootstrap from "uiBootstrap";

angular.app("myApp", [uiRouter, uiBootstrap])
```


This is mostly a moot point, as the 1.0 branch has embraced typescript.

Because of this, I'm closing this PR

PR #2065, opened 2015-06-26
fix(urlMatcherFactory): include the slash in squashed params

### LINKED ISSUE #2064: URL matcher with squash param is ignoring slashes
If I define a parameter in the url and state that squash is true for it, it will ignore slashes and think that things after the id.

I want `/test` to be the url for the state. However, if squash is true, then it will also match `/test123` and use 123 as the id. It should not match the state.

http://plnkr.co/edit/oIdIw3ntlqUpzHxhTWQe?p=preview
Example of the bug: http://run.plnkr.co/y1ii0hP2lnbQ3DLC/#/test123

```javascript
$stateProvider.state('layout', {
    url: '/test/:id',
    params: {
      id: {
        value: null,
        squash: true
      }
    },
...
```

eddiemonge:
Squashed params were ignoring a missing slash before the param so things
like /users123 would match /url/:id

Fixes #2064

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -103,7 +103,10 @@ function UrlMatcher(pattern, config, parentMatcher) {
     if (!pattern) return result;
     switch(squash) {
       case false: surroundPattern = ['(', ')' + (optional ? "?" : "")]; break;
-      case true:  surroundPattern = ['?(', ')?']; break;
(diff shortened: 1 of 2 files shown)


nateabele:
@eddiemonge Hey, just saw this. Should be good. I'll poke at it a bit later today.

PR #2081, opened 2015-07-05
fix (uiSrefActive): also update when stateParams change

### LINKED ISSUE #2049: ui-sref-active don't activate for queryParams + reloadOnSearch=false
Scenario:
The state defines `reloadOnSearch: false`
You interact with ui-sref directive that **only change query param** and stay in the same state.
The ui-sref-active directive **doesn't activate** the element because $stateChangeSuccess is never triggered.

Example:
http://plnkr.co/edit/cdaAnNXa2WkChlpAGbAq?p=preview

The [$StateRefActiveDirective](https://github.com/angular-ui/ui-router/blob/master/src/stateDirectives.js#L252) is made upon the `$stateChangeSuccess` event as the trigger to check if the element should be highlighted or not.

Here are some options I believe could be used to make ui-sref-active support query params change.

**Option 1**:
Include new event `$stateChangeSkipped` on the state transitioning livecycle that notifies when the current transition [was skipped](https://github.com/angular-ui/ui-router/blob/master/src/state.js#L1437) and then listen to it inside the directive to also check if need to activate the current elemen.

**Open 2**:
Trigger `$stateChangeSuccess` even if the transition was skipped.

Implementation wise both should be simple changes to do, I am curious to hear if people are experience same kind of problems and what are the thoughts about the proposed solution options. 

Related Issues:
#1752
#1675

thiagofelix:
- This include a new event on the state lifecycle named
  stateParamsChange that is fired inside the skipStateTransition
scenario.

Closes #2049

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -1042,6 +1042,25 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
           $urlRouter.update(true);
         }
         $state.transition = null;
+
+        // Broadcast params changed event
+        if (nonSearchParamsEqual(from, fromParams, toParams) && options.notify) {
+          /**
+           * @ngdoc event
+           * @name ui.router.state.$state#$stateParamsChange
+           * @eventOf ui.router.state.$state
+           * @eventType broadcast on root scope
+           * @description
+           * Fired during the state transition when state is the same but stateParams are different.
+           *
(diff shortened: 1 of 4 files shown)


nateabele:
We're not accepting PRs for any new events, sorry.

thiagofelix:
Fair enought, ther

PR #2086, opened 2015-07-08
fix(urlMatcherFactory.js): Url will not be matched if param value is not of param type and it is not optional

kruczy:
Hi,
 this is a simple patch to not match url if the parameter value is invalid and not optional I think it is necessary as not all custom types can have a regex that represents all valid values and if they don't router does not resolve subsequent routes properly


kruczy:
Any chance of getting this merged? or is there a better way to fix it? anyone?

nateabele:
I think this makes sense. @christopherthielen?

Also: http://prjs.radify.io/#/angular-ui/ui-router/pulls/2086

christopherthielen:
What if the parameter IS optional, but it doesn't match the `.is()` test?

kruczy:
If it is optional does it have to pass the test? Could remove the isOptional bit if it does

PR #2117, opened 2015-07-22
2110 add license to bower json

BernhardRode:
#2110 Add missing license in bower.json.


eddiemonge:
You need to squash your commits and update the commit message

PR #2119, opened 2015-07-22
fix(uiSrefActive): apply for all uiSref child elements

tiagoroldao:
For facilitating usage with menu elements with multiple children but no single parent, or cases where parent is not specified in parent.

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -253,18 +253,23 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
 
       // Update route state
       function update() {
-        for (var i = 0; i < states.length; i++) {
-          if (anyMatch(states[i].state, states[i].params)) {
-            addClass($element, activeClass);
-          } else {
-            removeClass($element, activeClass);
-          }
+        var matchedAny, matchedExact;
 
-          if (exactMatch(states[i].state, states[i].params)) {
-            addClass($element, activeEqClass);
-          } else {
-            removeClass($element, activeEqClass);
-          }
+        for (var i = 0; i < states.length; i++) {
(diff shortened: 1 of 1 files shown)


tiagoroldao:
I ran into problems while testing: basically, when testing, the aforementioned problem can't be replicated.

The following test passes prior to the change, when it shouldn't:
```javascript
it('should match on all child ui-sref directives', inject(function($rootScope, $q, $compile, $state) {
    el = angular.element('<div ui-sref-active-eq="active"><a ui-sref="contacts()">Contacts</a><a ui-sref="contacts.item({ id: 2 })">Contact</a><a ui-sref="contacts.item.detail({ id: 2 })">Contact</a></div>');
    template = $compile(el)($rootScope);
    $rootScope.$digest();

    expect(angular.element(template[0]).attr('class')).toBe('ng-scope');

    $state.transitionTo('contacts');
    $q.flush();
    timeoutFlush();
    expect(angular.element(template[0]).attr('class')).toBe('ng-scope active');

    $state.transitionTo('contacts.item', { id: 2 });
    $q.flush();
    timeoutFlush();
    expect(angular.element(template[0]).attr('class')).toBe('ng-scope active');

    $state.transitionTo('contacts.item.detail', { id: 2 });
    $q.flush();
    timeoutFlush();
    expect(angular.element(template[0]).attr('class')).toBe('ng-scope active');
}));
```

In practical usage it is seen to break (only matches the last ui-sref path). I experimented and saw that as long as one addClass is run by uiSrefActive, karma sees the class. Adding an unconditional removeClass at the end of the callback makes no difference - only calling removeClass directly on the test seems to make a difference.

tiagoroldao:
Apologies, went about fixing what's fixed. Was testing against 0.2.13, out of stupidity

PR #2149, opened 2015-08-03
feat($state): add prependBasePath option

williamboman:
Also s/appendBasePath/prependBasePath/g

So, I want to generate a non-absolute href without the base path prepended and couldn't find a way of doing that.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -1332,7 +1332,9 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
      * - **`relative`** - {object=$state.$current}, When transitioning with relative path (e.g '^'), 
      *    defines which state to be relative from.
      * - **`absolute`** - {boolean=false},  If true will generate an absolute url, e.g. "http://www.example.com/fullurl".
-     * 
+     * - **`prependBasePath`** - {boolean=true}, If false will not prepend base path (`<base href="/path/">`) to path. Will always prepend
+     *    base path if `absolute` is true.
+     *
      * @returns {string} compiled state url
      */
     $state.href = function href(stateOrName, params, options) {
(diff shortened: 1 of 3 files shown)


williamboman:
Status on this?

nateabele:
Hi, sorry for not responding earlier. I don't really understand the use case for this, and I don't see how it helps you do anything that you couldn't do by generating the HREF and concatenating `$("base").attr("href")` yourself.

williamboman:
@nateabele The default behaviour now is to concatenate the base path, with no option to turn it off. So in order to remove the base href, you'd have to do something like

```js
let href = $state.href(nextStateName, nextParams).replace(new RegExp(`^ + ${$('base').attr('href')}`));
```

to remove the base path.

So this patch makes it possible to opt-out of that behaviour (rather than opt-in - which is default).

nateabele:
Or better yet...

```javascript
let href = $state.href(nextStateName, nextParams).substr($('base').attr('href').length);
```

Again, though, what's the use case?

williamboman:
@nateabele Something like [this](http://plnkr.co/edit/7fbYOhgaytpDRjOfE6UV?p=preview).

nateabele:
Okay, that makes sense. However, I'm still pretty firmly against an option flag for every one-off use case that comes up, especially when they save very little code. People propose these constantly and it would quickly get out of hand. Sorry.

I am willing to consider exposing a simple API that would allow mapping a URL to a state + parameters, which would be a useful alternative in your case, in addition to being generally useful for many developers.

williamboman:
How about not prepending the base path if the href is not absolute? i.e.

```js
if (options.absolute) {
    url = prependBasePath(...);
}
```
It's essentially what this patch does, but in a BC manner. The native `$location` service omits the base path for relative paths.

nateabele:
I suppose because that would break Cmd/Ctrl-clicking or copying the URL?

williamboman:
Should it though? Browsers should take care of prepending any base path to relative links, e.g. `<a href>`, `<script src>`. I don't see why a non-absolute `$state.href()` should prepend the base path.

PR #2150, opened 2015-08-04
handling view refresh failure with try catch statement (was a known i…

tahir99:
…ssue on iPad ios8 Safari)


christopherthielen:
Thanks for the PR!  I'm going to have to close this for the following reasons:

- It could potentially cause an infinite loop
- It doesn't handle `firstTime == false`
- There are no tests

But most importantly:

- It looks to be a band aid for some other issue.  We would want to address the root issue, not hide it.

Is there a ticket on the issue tracker that describes the known issue on iPad ios8 Safari?

Issue #2146, opened 2015-07-31
resolve is always called with notify:false


davideicardi:
I want to change the state without reloading the controller. All seems to work fine except that my `resolve` function is always called.

To change state I use:

    $state.go("home", {param1:"changed" + changes++}, { notify : false});

Here my plnkr: http://plnkr.co/edit/jo3C56ZwkPE27FtRU1Y6

Basically I log each time the controller is loaded or the resolve is called, then I added a button to change the state. As you can see resolve is always called also with notify = false.

What do you think? My problem is that inside the resolve I load some resource and I want this code to be called only when really changing state, not when changing state using `notify=false`.

I use version ui-router 0.2.15 and angular 1.4.3.

eddiemonge:
Does seem like a bug

davideicardi:
Thinking more about this...`notify` is documented as: 

>  If true will broadcast $stateChangeStart and $stateChangeSuccess events

Documentation doesn't mention nothing about controller and resolve feature. 
My goal is to change the state (logically and also changing the location) without reloading the controller and without calling the resolve method again. Is the `notify:false` the right approach?

thanks

davideicardi:
Looking at the code I have a possible solution, here my fork+commit: https://github.com/davideicardi/ui-router/commit/7198aa2b3618f11b376949a80e2b7c4c6640cd4c

Basically I have checked the notify options inside shouldSkipReload function. I have also added a unit test. (don't know why there are also other CRLF changes...).
If you want I can create a pull request.

Eventually I have also a workaround, [http://plnkr.co/edit/hz3f7qORaHFZyiwHDZo7?p=preview](plunkr). I have added a custom "resolve" params and use it as an option to check if calling or not the resolve functions. So I can call the $state.go like this:

    $state.go("home", {param1:"changed" + changes++, resolve: false}, { notify : false});

But I don't like this solution, seems to be just a workaround.

What do you think? I'm not sure what is the right approach...

PhiLhoSoft:
Seems related to #2108 and #2087

eddiemonge:
#2108 has the controller reinitialize when changing to a different state. This has the resolve fired but not the controller so its slightly different.

nateabele:
> Documentation doesn't mention nothing about controller and resolve feature.

Right... that's because it's totally unrelated: `notify` is _only_ about events. Why is this confusing?

davideicardi:
[...]

Issue #2108, opened 2015-07-20
Controller initialised after a $state.go with {notify: false}


caillou:
## The Bug

When a `$state.go` is called with `{notify: false}` as follows, a bug occurred:

```javascript
$state.co('foo', {}, {notify: false})
```

On the **next state change**, the current controller will be initialised a second time.

## Plunker Link

Follow the steps to reproduce on Plunker:

http://plnkr.co/edit/wMiVOb?p=preview

eddiemonge:
It does indeed look like a bug.

PhiLhoSoft:
See #2087, looks similar. Very annoying when a controller loads data on initialization.

RickWalker:
This is the same issue as [#1673](https://github.com/angular-ui/ui-router/issues/1673) which is apparently fixed by PR [#2003](https://github.com/angular-ui/ui-router/pull/2003), so hopefully it should be easy enough to solve.

eddiemonge:
similar to #1673 but slightly different I think

jessezhang91:
It's a little different from #1673 but should still be fixed by PR #2003. The actual issue I was having is described in this bug.

eddiemonge:
marking as duplicate of #2087

Issue #2087, opened 2015-07-09
Notify:false fires the controller twice when changing state


PhiLhoSoft:
## The Bug

When a `$state.go` is called with `{notify: false}` as follows, a bug occurred:

```javascript
$state.go('foo', {}, {notify: false})
```

On the **next state change**, the current controller will be initialised a second time.

## Plunker Link

Follow the steps to reproduce on Plunker:

http://plnkr.co/edit/wMiVOb?p=preview
and
http://plnkr.co/edit/iqDuQrsjVWfySOrg1nv0?p=preview

----

Navigate to `foo` state, then click on *Change URL* button. The URL changes without reloading the controller, which is the wanted behavior.

Click on the *Change state 1* button. 
The URL changes, I see a `$stateChangeStart` event with the current URL parameters, the controller is called. 
Then I see a `$stateChangeSuccess` event with the new URL parameters and the controller is called again.

If you click now on the *Change state 2* button, both events have the same URL parameters, so there is no double controller call.

It looks like the `notify: false` parameter don't do a "real" URL parameters update, so UI-router has to do a double state change to reconcile the parameters it knows with the ones it discovers, then with the ones requested...

I saw a number of issues more or less related, like #1758, but I am not sure if they are entirely the same. So I went ahead and made my own issue with a demonstration to illustrate it. :smile:

PhiLhoSoft:
I feel this issue went unnoticed, as more recent ones are triaged as bug or design or similar, or get answers.
Might I have at least a comment on it? Is it a bug? Is there a workaround? Am I doing something wrong?

christopherthielen:
A few comments:

The docs for `notify` says: 
> If true will broadcast $stateChangeStart and $stateChangeSuccess events.
Using `notify: false` will simply suppress the firing of those events.  I realize that when this option was originally introduced, it was intended to implement the use cases ya'll are using it for.  However, the ramifications of suppressing those events was never fully thought out, and in fact it turned out to be a Very Bad Idea™ because it leaves the router internally inconsistent.  Therefore, please do not use `notify: false`.

[...]

PR #2154, opened 2015-08-05
Prevent running the resolve of the parent state when the user will navigate through child states

cssensei:
This code will prevent running the ``resolve`` of the parent state when the user will navigate through child states. (credits David Ficher - CustomerMatrix)

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -965,7 +965,9 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
         location: true, inherit: false, relative: null, notify: true, reload: false, $retry: false
       }, options || {});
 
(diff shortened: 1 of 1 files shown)


cssensei:
Might as well resolve these issues- https://github.com/angular-ui/ui-router/issues/2146, https://github.com/angular-ui/ui-router/issues/2108, https://github.com/angular-ui/ui-router/issues/2087

eddiemonge:
commit message doesn't follow the contributing guidelines. Also can you write tests showing what this is solving?

cssensei:
@eddiemonge I've made a mistake with this pull request. Sorry! It's actually best to close this one since it was just a bad use of ui-router from my side.

PR #1990, opened 2015-05-27
fix(angular-ui-router.d.ts): Delete angular-ui-router.d.ts

kenhowardpdx:
This file is out of date and breaks the build process. Deleting to defer to the community supported DefinitelyTyped version - https://github.com/borisyankov/DefinitelyTyped/issues/4035

PR #2180, opened 2015-08-17
Replace `declare ng.ui` with `declare angular.ui`.  Fixes #2035

### LINKED ISSUE #2035: Typescript definition problems
Hi! we have few errors when we integrate the definition file angular-ui-router.d.ts on our projects.
Here the errors:

```
Import declaration conflicts with local declaration of 'ng'	
Cannot find name 'IServiceProvider'.
Cannot find name 'IPromise'.
```

I think namespaces are changed...my two cents.

rcollette:
Because the TS 1.5 language service does not yet support the exclude option, IDEs such as WebStorm will find and use angular-ui-router.d.ts.

As discussed in issue 2035, the declaration ng.ui conflicts with the declaration of ng in angular.d.ts.

This merge is to correct the declaration from ng.ui to angular.ui

Pull 1990 is attempting to address the issue by removing the definition entirely from the project, which is fine as well.

Either way, the issue is a blocker for those compiling with an IDE (i.e. perhaps not using  a gulpfile).

Hopefully one of the two pulls will be merged ASAP.

Fixes #2035

### DIFF
--- a/api/angular-ui-router.d.ts
+++ b/api/angular-ui-router.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Michel Salib <https://github.com/michelsalib>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
 
-declare module ng.ui {
+declare angular.ui {
 
     interface IState {
         name?: string;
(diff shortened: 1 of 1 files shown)


wesleycho:
This should be addressed with 63f4476 from #1990. Please use the implementation in https://github.com/borisyankov/definitelytyped

PR #2187, opened 2015-08-19
feat(uiState): add ui-state directive

### LINKED ISSUE #395: Dynamic ui-sref in directive

nateabele:
- Refactor StateRefDirective for better modularity
 - Drop key restrictions on ui-sref-opts
 - Improves performance over prior implementation with no extra $eval()’s

Fixes #395, #900, #1932

cc: @kasperlewau @christopherthielen

### DIFF
--- a/test/stateDirectivesSpec.js
+++ b/test/stateDirectivesSpec.js
@@ -291,6 +291,71 @@ describe('uiStateRef', function() {
     }));
   });
 
+  describe('links with dynamic state definitions', function () {
+    var template;
+
+    beforeEach(inject(function($rootScope, $compile, $state) {
+      el = angular.element('<a ui-state="state" ui-state-params="params">state</a>');
+      scope = $rootScope;
+      angular.extend(scope, { state: 'contacts', params: {} });
+      template = $compile(el)(scope);
+      scope.$digest();
+    }));
+
+    it('sets the correct initial href', function () {
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts');
+    });
+
+    it('updates to the new href', function () {
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts');
+
+      scope.state = 'contacts.item';
+      scope.params = { id: 5 };
+      scope.$digest();
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts/5');
+
+      scope.params.id = 25;
+      scope.$digest();
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts/25');
+    });
+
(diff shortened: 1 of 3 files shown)


test/stateDirectivesSpec.js
```diff
@@ -291,6 +291,71 @@ describe('uiStateRef', function() {
...
+
+    it('sets the correct initial href', function () {
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts');
+    });
+
+    it('updates to the new href', function () {
+      expect(angular.element(template[0]).attr('href')).toBe('#/contacts');
+
+      scope.state = 'contacts.item';
+      scope.params = { id: 5 };
```
kasperlewau: This looks way better to me than what I previously had; https://github.com/kasperlewau/ui-router/commit/110d793ee6591b139d390f49c642edc93326fe74#diff-e571eaac1af787a06dfe6dfe7210a4d9R312. 

Specifying the state ref *including* the params as a string felt sort of wonky when all things come around. Makes more sense to split it. :+1:

christopherthielen:
:+1:

christopherthielen:
:+1:

jasdeepsingh:
:+1: needed this. thanks ;)

vinothbabu:
Was looking for this... thanks guys...

mrmashal:
Thanks for the dynamic ui-sref, but:
1. How it should be used? Is it: `<a ui-state="state" ui-state-params="params" ui-state-opts="{ absolute: true }"></a>`?
2. Does it support passing state parameters that are not used in the URL?

Travistom:
Thanks

PR #2198, opened 2015-08-24
fix(UrlMatcherFactory): Correct trailing slash of terminal optional parameter

### LINKED ISSUE #1902: Slash is not removed for squash:true when optional param terminates the url
In @christopherthielen's [plunkr](http://plnkr.co/edit/VMPc8D7oUG0B1R3QuiCE?p=preview) example from #1501, `:username`'s slash is not removed when it is squashed. The documented behaviour for `squash: true` is that the optional parameter plus a slash are removed.

Reprod:
 1. Open the plunkr
 1. Click on the purple `user({username: 'christopherthielen'})` button.

Observed output: `/user/`
Expect output: `/user`

jshado1:
Fixes #1902

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -342,6 +342,8 @@ UrlMatcher.prototype.format = function (values) {
 
     if (isPathParam) {
       var nextSegment = segments[i + 1];
+      var isFinalPathParam = i + 1 === nPath;
+
       if (squash === false) {
         if (encoded != null) {
           if (isArray(encoded)) {
@@ -357,6 +359,8 @@ UrlMatcher.prototype.format = function (values) {
       } else if (isString(squash)) {
         result += squash + nextSegment;
       }
+
+      if (isFinalPathParam && param.squash === true && result.slice(-1) === '/') result = result.slice(0, -1);
     } else {
       if (encoded == null || (isDefaultValue && squash !== false)) continue;
       if (!isArray(encoded)) encoded = [ encoded ];
--- a/test/urlMatcherFactorySpec.js
+++ b/test/urlMatcherFactorySpec.js
@@ -149,6 +149,14 @@ describe("UrlMatcher", function () {
           params = { id: 'bob', section: 'contact-details' };
       expect(m.format(params)).toEqual('/users/bob#contact-details');
     });
+
+    it("should trim trailing slashes when the terminal value is optional", function () {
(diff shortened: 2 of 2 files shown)


christopherthielen:
Thanks jshado1

Does this perform symmetrically, i.e., does it also match properly when the trailing slashes are omitted?
Also, I don't think this obeys the $urlMatcherFactory.strictMode setting.  

Can you address those issues and push an updated PR?

I'd like to see additional unit tests for more combinations, with the above questions in mind

src/urlMatcherFactory.js
```diff
@@ -342,6 +342,8 @@ UrlMatcher.prototype.format = function (values) {
 
     if (isPathParam) {
       var nextSegment = segments[i + 1];
+      var isFinalPathParam = i + 1 >= nPath;
```
jshado1: this should probably be changed to `i + 1 === nPath` (`isPathParam` already ensures that this is within `nPath`).

jshado1:
Hi @christopherthielen,

I think there're other tests that (inadvertently?) assert cases 1 and 2. I'll look for them.

Is it safe to assume that [param.squash](https://github.com/jshado1/ui-router/blob/1902/src/urlMatcherFactory.js#L340) will inherit `defaultSquashPolicy` when it hasn't been explicitly set?

Thanks for the feedback.

nateabele:
> Also, I don't think this obeys the $urlMatcherFactory.strictMode setting.

That should really only apply to parsing.

> Is it safe to assume that param.squash will inherit defaultSquashPolicy when it hasn't been explicitly set?

https://github.com/jshado1/ui-router/blob/1902/src/urlMatcherFactory.js#L202

@jshado1 If you wanna make the change you noted above and squash your commits, this is :+1: from me.

jshado1:
@christopherthielen, non-optional is covered by [urlMatcherFactorySpec.js#L617](https://github.com/angular-ui/ui-router/blob/master/test/urlMatcherFactorySpec.js#L617).

> Does this perform symmetrically, i.e., does it also match properly when the trailing slashes are omitted?

Possibly lacking caffeine, but what is "it" ?

jshado1:
@nateabele I made that change I had commented on (and squashed).

christopherthielen:
Regarding my symmetry comment, I want to make sure that if strict mode is on and a url is generated (with trailing slash removed), that the url will then be matched correctly to the same state and paeans (for instance if the user hits reload in the browser)

PR #2201, opened 2015-08-24
feat($state): Added default option for $state.go

a60814billy:
- added 'defaults' property to $stateProvider
- modified $state.transitionTo default options to $stateProvider defaults property

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -688,6 +688,32 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
 
   /**
    * @ngdoc object
+   * @name ui.router.state.$stateProvider
+   *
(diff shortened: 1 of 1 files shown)


christopherthielen:
in 1.0, the default transition options are now in the transition service, so I don't think we should add them to the $state service.  

https://github.com/angular-ui/ui-router/blob/feature-1.0/src/transition/transitionService.ts#L30-L39

nateabele:
Agreed. This is a decent idea, but we don't want to introduce any more changes that would cause 0.2.x to diverge any further from 1.0.

a60814billy:
ok, thanks for yours reply. I will be using 1.0 version soon.

PR #2218, opened 2015-08-31
feat(urlMatcher): add raw encoding flag to Type

### LINKED ISSUE #1119: url encode stateparams

nateabele:
Closes #1119

### DIFF
--- a/src/params/param.ts
+++ b/src/params/param.ts
@@ -1,4 +1,4 @@
-import {isInjectable, extend, isDefined, isString, isArray, filter, map, prop} from "../common/common";
+import {isInjectable, extend, isDefined, isString, isArray, filter, map, prop, curry} from "../common/common";
 import {runtime} from "../common/angular1";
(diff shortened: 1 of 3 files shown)


christopherthielen:
very nice.

mismith:
awesome work, thanks!

plong0:
will this be applied to the 0.2.* release branch too?  I just added support for a raw flag to type in my local 0.2.15. Should I make a PR for that?

nateabele:
Not sure yet. We're focused pretty heavily on getting 1.0 out the door, and since 1.0 is mostly feature-compatible with 0.2.x, getting everyone to upgrade.

PR #2226, opened 2015-09-03
fix(urlRouter): add check using `$sniffer`

wesleycho:
- Add additional check for situations where `pushState` is not properly supported


wesleycho:
Need a review here as to what is going on with the tests - I'm puzzled.

test/urlRouterSpec.js
```diff
@@ -1,7 +1,6 @@
 describe("UrlRouter", function () {
 
-  var $urp, $lp, $ur, location, match, scope;
```
christopherthielen: You deleted these other variables accidentally

PR #2241, opened 2015-09-14
Improved $state.go documentation

tamlyn:
I spent a while stepping through the code to work out why the extra parameters I was passing to `$state.go()` weren't appearing in `$stateParams`. I've added a sentence to clarify why that is.


nateabele:
> "This allows, for example, going to a sibling state that shares parameters defined parameters."

Looks like some language needs to be massaged a bit.

tamlyn:
Thanks, corrected.

eddiemonge:
are you going to squash the commits?

nateabele:
For reference: http://prjs.radify.io/#/angular-ui/ui-router/pulls/2241

tamlyn:
Thanks, will do. My parents brought me up not to rewrite public history but I guess if no one has pulled it it's OK.

PR #2254, opened 2015-09-20
Fix incorrect name of transitionsProvider on readme

charandas:
Its currently `transitionProvider` in the readme which is not what it is actually :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,24 +30,24 @@ When registering a hook, you can provide criteria (a state name, a glob, or a fu
 
 This enables lots of fun stuff!  Here are a couple of possibilities to get your imagination started:
 ```
-$transitionProvider.onBefore({ to: 'my.state', from: '*' }, function(AsyncService) {
+$transitionsProvider.onBefore({ to: 'my.state', from: '*' }, function(AsyncService) {
   return AsyncService.doSomeAsyncThing();
 });
 
-$transitionProvider.onBefore({ to: 'other.state', from: '*' }, function(AsyncService) {
+$transitionsProvider.onBefore({ to: 'other.state', from: '*' }, function(AsyncService) {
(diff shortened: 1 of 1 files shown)


christopherthielen:
@charandas are you playing with feature-1.0 codebase?  How are you finding it?

charandas:
@christopherthielen Thanks for checking in. Yes, I am interested in updating to it.

At first glance, couldn't have play well with ui-router-extras. There was some issue with a decorator call in the library.

Trying to see if my app can instead run initialization code in transition hooks.

charandas:
Excited about what you guys are targeting for near future, for Angular 2 especially.

christopherthielen:
@charandas ui-router-extras sticky states is not ready for the 1.0 branch.  
future states likely works.  
DSR and previous states likely do NOT work (without enabling the now-deprecated-state-events in the 1.0 $stateEventsProvider)

charandas:
Yeah, cannot expect you to be on top of that project alongside the gargantuan duties that come with being a ui-router maintainer :grinning: 

I am interested in using the dynamic params (aka $location.search()) without a state reload with 1.0 preview. I read somewhere there is a decorator for that. Care to enlighten, please?

christopherthielen:
Params in general (and dynamic params) are one of the chores on my list to address before the 1.0-beta.    We currently have three concepts internally managing params (ParamSet, StateParams, ParamValues) and I need to consolidate them down to two, or perhaps even one class.  As of right now, dynamic params *might* work if you use $state.params global mutable state object.  

However, we're moving people *away* from using global mutable state.  I want people to watch dynamic params using the $stateParams object injected into a hook or controller, but dynamic params aren't currently implemented there.

charandas:
Thanks for the update. Will stay tuned. :+1:  @christopherthielen

PR #2269, opened 2015-09-28
feat: support params in isState filter

philBrown:


nateabele:
Patch & test look good. Please see http://prjs.radify.io/#/angular-ui/ui-router/pulls/2269.

philBrown:
@nateabele all fixed

PR #2279, opened 2015-10-01
feat($state): make state data inheritance prototypical

chrislambe:
Modify state data inheritance to use prototypical inhertiance rather than Angular's extend method to enable more flexibility in handling state data.

### Justification
I'm currently working on a project that implements Material Design via Angular Material. We've implemented a global [FAB](https://www.google.com/design/spec/components/buttons-floating-action-button.html) by attaching information about what the FAB should do to the state definition. For example:

```javascript
angular.module('MyMaterialDesignProject')
  .config(function($stateProvider) {
    $stateProvider
      .state('foo', {
        url: '/foo',
        data: {
          primaryAction: {
            onClick: function($state) {
              $state.go('foo.bar');
            }
          }
        }
      })
      .state('foo.bar', {
        url: '/bar'
      });
  });
```

When we arrive at `foo.bar` we need to be able to tell if `data.primaryAction` is defined as part of the `foo.bar` state definition or if it's inherited. If it's inherited we (maybe) don't want to display the FAB.

### Risk of breaking change
Any dev who has implemented a test to determine the availability of a data property using `data.hasOwnProperty('foo')` rather than something like `'foo' in data` is at risk of having their app stop functioning as they expect.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -41,7 +41,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
     // inherit 'data' from parent and override by own values (if any)
     data: function(state) {
       if (state.parent && state.parent.data) {
-        state.data = state.self.data = extend({}, state.parent.data, state.data);
+        state.data = state.self.data = inherit(state.parent.data, state.data);
       }
       return state.data;
     },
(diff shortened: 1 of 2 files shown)


nateabele:
@christopherthielen Obviously we'll want to use `inherit()` but I am fundamentally down with this. Any objection?

chrislambe:
Switched to using `inherit()`.

christopherthielen:
I totally thought we were already doing it this way, hah.  :+1: 

@chrislambe
FYI I use `data` for metadata the entire tree of states cares about, but I've been plopping non-inherited metadata directly on the state declarations, e.g.,
```
.state('admin', {
 url: '/admin',
 dsr: true, // <- plop
 data: {
  roles: ['ADMIN']
 } 
}```

chrislambe:
@christopherthielen Oh, interesting. That's helpful, thanks!

PR #2283, opened 2015-10-03
Add state alias

Melanitski:
If we have complicated nested route name, we can use `alias`
Example:

```
  .state 'public.search',
    alias : 'search'
    url: 'search'
    template: 'views/search'

# Usage
ui-sref="search"
$state.go("search")
```


eddiemonge:
Why not:

```
.state 'search',
  parent: 'public'
```
?

bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "angular-ui-router",
-  "version": "0.2.15",
+  "version": "0.2.16",
```
eddiemonge: this happens as part of the build process so should not be in your commit

eddiemonge:
for pull requests like this, only include the actual files related to your change, not the build related things like version changes or updated dist code. Also, one commit per pull request. Also, you should follow the [contributing guide](https://github.com/angular-ui/ui-router/blob/master/CONTRIBUTING.md)

Melanitski:
If we have very complicated router like 'public.search.item.compare' we can make alias like a `compare` or `items-compare` 
We need to add tests and dist files?

eddiemonge:
```
state: 'compare',
  parent: 'public.search.item'
```

christopherthielen:
I agree with @eddiemonge . While I am not fundamentally opposed to an alias, I think we already have a mechanism that provides the desired behavior (`parent: `).  

Also, ui-sref supports relative addressing, so within a template for `public.search.item`, you simply enter a link `<a ui-sref=".compare">Compare</a>`.  Likewise, `$state.go()` also accepts relative state names.

Closing for now; Feel free to respond with additional discussion.

Issue #983, opened 2014-03-27
Can we pass objects these days with transitionTo ?


Bretto:
can we pass objects these days with transitionTo ?

kilaka:
Tried and worked using `transitionTo` . Don't know if works with `go`.

shavo007:
:+1:

filipesilva:
Tried and worked with `go`, but I'm getting some inconsistent behavior depending of the parameter.

The target state seems to only reloaded if the param is different at a shallow level. Example in [this plunker](http://plnkr.co/edit/JLVk4rvwpAU3AM7TTxXO?p=preview).

In that example, I have an array that is filled with random numbers when you click the link. 

If I send that array into `state.go` directly, the child state will not be reloaded (check console for `child loaded` messages). But if I `slice` the array (creating copy of the array reference) then the child state is reloaded.

I would expect the state to be reloaded regardless, since the contents of the parameter changed.

christopherthielen:
I'm pretty sure we don't do deep comparisons by default. I think we do 'a == b' and since you reuse the same array reference they are '=='.  We have support for array-type for url parameters, but I'm not sure how that would work for non URL params ('array: true'). Otherwise, implement your own type with an equals() that deep compares.  Come to think of it, you might be able to set the param type to "json" which does deep compare.

filipesilva:
I can't seem to find the param type 'json' in the documentation, is it a new addition?

christopherthielen:
Your plunk works with 'json' type:   http://plnkr.co/edit/kBmbxyaU8eWVlNKtEHwX?p=preview

Yeah, json parameter type was added here: https://github.com/angular-ui/ui-router/releases/tag/0.2.13  but looks like it never got documented

christopherthielen:
As a followup, I actually created the "any" type for exactly your scenario.  It doesn't encode/decode and uses angular.equals to do a deep compare.  Use that instead of 'json' type.  

However, I also thought that non-url parameters were automatically declared as "any" types, hrm...

filipesilva:
Indeed types 'json' and 'any' solve my issue on my production website, many thanks!

Going over those commits, it seems the whole 'type' section of params is undocumented [here](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$stateProvider).

Would a PR of mine to document it be welcome, or is this something on the pipeline for later on?

christopherthielen:
[...]

PR #2289, opened 2015-10-06
feat(UrlMatcher): Add param only type names

filipesilva:
As per https://github.com/angular-ui/ui-router/issues/983#issuecomment-145288637

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -917,7 +917,8 @@ function $UrlMatcherFactory() {
     }
 
     function getType(config, urlType, location) {
-      if (config.type && urlType) throw new Error("Param '"+id+"' has two type configurations.");
+      if (config.type && urlType.name !== 'string') throw new Error("Param '"+id+"' has two type configurations.");
+      if (config.type && urlType.name === 'string' && $types[config.type]) return $types[config.type];
       if (urlType) return urlType;
       if (!config.type) return (location === "config" ? $types.any : $types.string);
       return config.type instanceof Type ? config.type : new Type(config.type);
--- a/test/urlMatcherFactorySpec.js
+++ b/test/urlMatcherFactorySpec.js
@@ -520,6 +520,21 @@ describe("urlMatcherFactory", function () {
       expect(m.format({ foo: 5, flag: true })).toBe("/5/1");
     });
 
+    it("should match types named only in params", function () {
+      var m = new UrlMatcher("/{foo}/{flag}", {
+        params: {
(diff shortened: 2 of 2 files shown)


src/urlMatcherFactory.js
```diff
@@ -917,7 +917,8 @@ function $UrlMatcherFactory() {
     }
 
     function getType(config, urlType, location) {
-      if (config.type && urlType) throw new Error("Param '"+id+"' has two type configurations.");
+      if (config.type && urlType.name !== 'string') throw new Error("Param '"+id+"' has two type configurations.");
+      if (config.type && urlType.name === 'string' && $types[config.type]) return $types[config.type];
```
christopherthielen: Since `urlType` is optional, this code will break on url-less parameters.  I liked the direction of your previous logic better, where it checks if config.type is a string.

`angular.isString` is a handy utility for checking if something is a string or not

src/urlMatcherFactory.js
```diff
@@ -917,7 +917,8 @@ function $UrlMatcherFactory() {
...
       if (urlType) return urlType;
       if (!config.type) return (location === "config" ? $types.any : $types.string);
       return config.type instanceof Type ? config.type : new Type(config.type);
```
christopherthielen: something like this:
```
if (angular.isString(config.type))
  return $types[config.type];
if (config.type instanceof Type)
  return config.type;
return new Type(config.type);
```

test/urlMatcherFactorySpec.js
```diff
@@ -520,6 +520,21 @@ describe("urlMatcherFactory", function () {
...
+          },
+          flag: {
+            type: 'bool'
+          }
+        }
+      });
+      expect(m.exec("/1138/1")).toEqual({ foo: 1138, flag: true });
+      expect(m.format({ foo: 5, flag: true })).toBe("/5/1");
+    });
+
```
christopherthielen: Please add a test for url-less parameters, like our discussion in issue #983

christopherthielen:
I see why you made the changes you did.  There are some interactions with the url param type.  I'm going to take this and massage it a bit.  Thanks for your efforts!

christopherthielen:
updated PR: https://github.com/angular-ui/ui-router/pull/2296

test/urlMatcherFactorySpec.js
filipesilva: I got so caught up on the type tests in the urlMatcher that I forgot tests for url-less params. My bad!

PR #2293, opened 2015-10-06
Revert "feat(UrlMatcher): Add param only type names"

nateabele:
Reverts angular-ui/ui-router#2289

After talking with @christopherthielen, this likely breaks cases that we realized we don't have test coverage for, so that needs to be fixed before this can be addressed. Sorry for the confusion.


nateabele:
cc: filipesilva

filipesilva:
No worries at all!

PR #2296, opened 2015-10-06
fix(urlMatcherFactory): fix configuring a parameter type by name in a…

### LINKED ISSUE #2294: Parameter types cannot be declared in the param config block
See discussion in #983  and original PR #2289

This should work, but does not:
```
.state({
 name: 'fooState',
 url: '/{param1:string}/{param2}',
 params: {
  param2: {
    value: 100, // default val
    type: 'int' // Setting a url param type here doesn't work
  },
  param3: {
    type: 'int' // url-less parameter type by name doesn't work
  }
 }
})
```

christopherthielen:
… params block

closes #2294

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -122,7 +122,11 @@ function UrlMatcher(pattern, config, parentMatcher) {
     cfg         = config.params[id];
     segment     = pattern.substring(last, m.index);
     regexp      = isSearch ? m[4] : m[4] || (m[1] == '*' ? '.*' : null);
-    type        = $$UMFP.type(regexp || "string") || inherit($$UMFP.type("string"), { pattern: new RegExp(regexp, config.caseInsensitive ? 'i' : undefined) });
+
+    if (regexp) {
+      type      = $$UMFP.type(regexp) || inherit($$UMFP.type("string"), { pattern: new RegExp(regexp, config.caseInsensitive ? 'i' : undefined) });
+    }
+
     return {
       id: id, regexp: regexp, segment: segment, type: type, cfg: cfg
     };
@@ -920,7 +924,12 @@ function $UrlMatcherFactory() {
       if (config.type && urlType) throw new Error("Param '"+id+"' has two type configurations.");
       if (urlType) return urlType;
       if (!config.type) return (location === "config" ? $types.any : $types.string);
-      return config.type instanceof Type ? config.type : new Type(config.type);
+
+      if (angular.isString(config.type))
+        return $types[config.type];
+      if (config.type instanceof Type)
+        return config.type;
+      return new Type(config.type);
     }
 
     // array config: param name (param[]) overrides default settings.  explicit config overrides param name.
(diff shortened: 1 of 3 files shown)


src/urlMatcherFactory.js
```diff
@@ -917,10 +921,15 @@ function $UrlMatcherFactory() {
     }
 
     function getType(config, urlType, location) {
-      if (config.type && urlType) throw new Error("Param '"+id+"' has two type configurations.");
+      if (config.type && urlType && config.type !== urlType) throw new Error("Param '"+id+"' has two type configurations.");
```
christopherthielen: whops, missed that.  I will fix and force push

src/urlMatcherFactory.js
```diff
@@ -122,7 +122,11 @@ function UrlMatcher(pattern, config, parentMatcher) {
     cfg         = config.params[id];
     segment     = pattern.substring(last, m.index);
     regexp      = isSearch ? m[4] : m[4] || (m[1] == '*' ? '.*' : null);
-    type        = $$UMFP.type(regexp || "string") || inherit($$UMFP.type("string"), { pattern: new RegExp(regexp, config.caseInsensitive ? 'i' : undefined) });
```
christopherthielen: defer applying the default path/query param type of "string" until the `getType` function.

src/urlMatcherFactory.js
```diff
@@ -920,7 +924,12 @@ function $UrlMatcherFactory() {
       if (config.type && urlType) throw new Error("Param '"+id+"' has two type configurations.");
       if (urlType) return urlType;
       if (!config.type) return (location === "config" ? $types.any : $types.string);
```
christopherthielen: Heres where we were applying default type 'string' to url params (Actually, to non-non-url params ;) )

test/stateSpec.js
```diff
@@ -32,6 +32,7 @@ describe('state', function () {
       RSP = { url: '^/:doReload/search?term', reloadOnSearch: false },
       OPT = { url: '/opt/:param', params: { param: "100" } },
       OPT2 = { url: '/opt2/:param2/:param3', params: { param3: "300", param4: "400" } },
+      URLLESS = { url: '/urllessparams', params: { myparam: { type: 'int' } } },
```
christopherthielen: non-optional non-url param of type `int`

src/urlMatcherFactory.js
filipesilva: Yes this was the bit that was setting things as type string for me. I didn't fully understand all the code so I tried to confine my changes to `getType` as to limit impact.

test/urlMatcherFactorySpec.js
```diff
@@ -520,6 +520,30 @@ describe("urlMatcherFactory", function () {
...
+          foo: { type: 'int'},
+          flag: { type: 'bool'}
+        }
+      });
+      expect(m.exec("/1138/1")).toEqual({foo: 1138, flag: true});
+      expect(m.format({foo: 5, flag: true})).toBe("/5/1");
+    });
+
+    it("should throw an error if a param type is declared twice", function () {
+      try {
```
nateabele: Don't you want `expect(...).toThrow(...)`?

christopherthielen: updated

nateabele:
This looks :+1: to me, and should translate 1:1 onto the params branch.

PR #2297, opened 2015-10-07
feat(stateFilters): create filter handling state params

benjamincanac:


src/stateFilters.js
```diff
@@ -31,9 +31,28 @@ function $IncludedByStateFilter($state) {
...
+/**
+ * @ngdoc filter
+ * @name ui.router.state.filter:includedByStateWithParams
+ *
+ * @requires ui.router.state.$state
+ *
+ * @description
+ * Translates to {@link ui.router.state.$state#methods_includes $state.includes('fullOrPartialStateName', {})}.
+ */
+$IncludedByStateWithParamsFilter.$inject = ['$state'];
```
wesleycho: No reason for a whole new filter - you just can extend the original filter by adding a second `params` argument.

benjamincanac:
You are totally right. See PR #2350

PR #2322, opened 2015-10-21
Throws an error if the invocation of a resolve function fails

BaggersIO:
Thrown errors from `$injector.invoke` get lost at https://github.com/angular-ui/ui-router/blob/master/src/resolve.js#L174.

This pull request will re-throw the error for better debugging. Otherwise you'll get lost in debugging broken resolve functions :)

In my case, Angular could not create a service properly. The state didn't resolved without an error.

### DIFF
--- a/src/resolve.js
+++ b/src/resolve.js
@@ -111,6 +111,7 @@ function $Resolve(  $q,    $injector) {
       function fail(reason) {
         result.$$failure = reason;
         resolution.reject(reason);
+        throw reason;
       }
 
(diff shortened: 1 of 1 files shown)


nateabele:
This has already been addressed elsewhere. Throwing errors breaks the promise abstraction. See `$stateChangeError` if you need help debugging. There will be a default error handler in 1.0 that makes this more obvious.

BaggersIO:
+1 :)

PR #2328, opened 2015-10-22
feat($state): add $stateChangeValidation event (rebased)

alxndrsn:
This is a clean up of the PR at https://github.com/angular-ui/ui-router/pull/1845, rebased to the latest `master`.

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -1155,7 +1155,7 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
            * })
            * </pre>
            */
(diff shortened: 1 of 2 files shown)


christopherthielen:
Thank you for your PR!  I appreciate the effort you have put in to create this and keep submitting it.

However, I don't think it aligns with our vision for ui-router's future.  I think the transition hooks in 1.0 fully address the async validation use case, and we have deprecated (and removed by default) the state events.  

The original PR #1845 mentions angular-permission as a driver; I see that project has started a branch which utilizes the 1.0 transition hooks.

PR #2346, opened 2015-10-30
$destroy event is triggered before animation ends

### LINKED ISSUE #1643: $destroy event is triggered on leaving view before animation/transition ends
When animations are associated to a view using ngAnimate, $destroy event is triggered on leaving view before animation/transition ends.

This event should be triggered after animation/transition end?

[Plunker](http://plnkr.co/edit/UQcrIpf2V44pXTROQmVm?p=preview)

angular@1.3.6
angular-animate@1.3.6
ui-router@0.2.13

Frozenball:
closes #1643

### DIFF
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -190,24 +190,37 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate)
         updateView(true);
 
         function cleanupLastView() {
-          if (previousEl) {
-            previousEl.remove();
-            previousEl = null;
+          var _previousEl = previousEl;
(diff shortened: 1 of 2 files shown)


nateabele:
@Frozenball Pro.

christopherthielen:
@Frozenball @ghoullier @philBrown  @Smulis @petermolyneux 

We're going to revert this PR, as deferring the scope destruction has caused all sorts of lifecycle issues.  

Can you pull/build PR #2596 and try using the `$animLeave` promise from `$element.data('$uiView')` instead?  

Let me know if the animation promise will satisfy your use cases.

PR #2350, opened 2015-11-02
feat(includedByStateFilter): Include optional state params.

benjamincanac:
Allows state params to be passed to the `includedByState` filter for more accurate matching.

### DIFF
--- a/src/stateFilters.js
+++ b/src/stateFilters.js
@@ -27,8 +27,8 @@ function $IsStateFilter($state) {
  */
 $IncludedByStateFilter.$inject = ['$state'];
 function $IncludedByStateFilter($state) {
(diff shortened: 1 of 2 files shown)


christopherthielen:
Thanks for the PR with tests!

I'm finding this pretty awkward: `ng-if="'somestate' | includedByState: { param: 'a' }'"`

`ui-sref` uses a DSL to include state and parameters, like `"somestate({ param: 'a' })`, but even `ng-if="'somestate({ param: 'a' }) | includedByState"` seems only marginally better. 

The ui-router team has discussed deprecating and eventually removing the state filters altogether (or packaging them separately).  Can you show me how you use this in your application?

christopherthielen:
See also discussion in #1947

PR #2363, opened 2015-11-08
feat(uiSrefActive): provide a ng-{class,style} like interface

### LINKED ISSUE #1431: ui-sref-active not working correctly with abstract and child states
http://plnkr.co/edit/c7OS1pwI5IjAAs5cEi5s?p=preview

In the example above, clicking the Administration nav element shows a view with two tabs, Users and Roles.

These correspond to the states `admin.users` and `admin.roles`, while `admin` is an abstract state whose URL is inherited by `admin.users`. This all works fine except when setting the active class using `ui-sref-active`.

There are two problems in this scenario:

1. The Administration nav element is not made active when on the Roles tab
2. When refreshing the page (be sure to pop out the Plunker preview) while on the Roles tab, no elements get the active class set

I would like to avoid using `$state.includes` directly since `ui-sref-active` is supposed to work on child states since 0.2.11: https://github.com/angular-ui/ui-router/pull/927

Am I doing something wrong or is this a bug?

fpipita:
This pull request is based on the suggestion by @nateabele on issue #1431. It enables `ui-sref-active` to be passed an expression that should evaluate to an object hash with the `activeClass` and `state` properties. If `state` is provided, `activeClass` is applied to its element if the current active state includes `state`, even if `state` is abstract. It works even if `ui-sref` is found neither on the element nor on any of its children. Furthermore, if `state` is provided, it takes precedence over the state exposed by `ui-sref`. Hope it helps to close #1431 :)

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -9,7 +9,8 @@ var isDefined = angular.isDefined,
     isArray = angular.isArray,
     forEach = angular.forEach,
     extend = angular.extend,
-    copy = angular.copy;
+    copy = angular.copy,
+    toJson = angular.toJson;
 
 function inherit(parent, extra) {
   return extend(new (extend(function() {}, { prototype: parent }))(), extra);
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -249,6 +249,24 @@ function $StateRefDynamicDirective($state, $timeout) {
  *   </li>
  * </ul>
  * </pre>
+ *
+ * It is also possible to pass ui-sref-active an expression that evaluates
+ * to an object hash, whose keys represent active class names and whose
+ * values represent the respective state names/globs.
+ * ui-sref-active will match if the current active state **includes** any of
+ * the specified state names/globs, even the abstract ones.
+ *
+ * @Example
+ * Given the following template, with "admin" being an abstract state:
+ * <pre>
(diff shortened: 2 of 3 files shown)


nateabele:
Hey, sorry for not responding right away. After a belated internal discussion with @christopherthielen, we agreed an `ng-class`-style interface would be better, i.e. with class names as keys and state names/globs as the values, i.e. `{ active: "admin.*" }` — I think that would provide a lot more flexibility and a lot less verbosity?

Would you be willing to update the PR to support that? If not, no worries, and sorry for the back-and-forth.

fpipita:
Hi, no problem at all about that! I definitely agree with you guys about the ng-{class,style} providing a cleaner interface.

I will update the PR in this week-end.

nateabele:
Awesome, thanks!

fpipita:
Hi, I've updated the PR to match the new requirements. Please, let me know if anything else needs to be done :)

christopherthielen:
Can we get another PR to merge to `feature-1.0`?

fpipita:
Hi, I've just opened PR #2375 which brings this feature into `feature-1.0`.

shali3:
This is a great feature. you should really document it. I see no mention of this.

maybenull:
Agree with @shali3, docs are missing this.

PR #2375, opened 2015-11-17
feat(uiSrefActive): provide a ng-{class,style} like interface

### LINKED ISSUE #1431: ui-sref-active not working correctly with abstract and child states

fpipita:
Ported the PR that resolves #1431 to the `feature-1.0` branch.

### DIFF
--- a/src/state/stateDirectives.ts
+++ b/src/state/stateDirectives.ts
@@ -1,5 +1,5 @@
 /// <reference path='../../typings/angularjs/angular.d.ts' />
-import {copy, defaults} from "../common/common";
+import {copy, defaults, isString, isObject, forEach, toJson} from "../common/common";
 import {defaultTransOpts} from "../transition/transitionService";
 
 function parseStateRef(ref, current) {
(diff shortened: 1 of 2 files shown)


christopherthielen:
I'll just assume this works.  That's good software dev, right?

PR #2383, opened 2015-11-20
feat($stateSearch): show $location.search for current state

kindy:
try to resolve #115.

we want to support dynamic search params.
but seems $stateSearch can help us too.


christopherthielen:
Thanks for your PR! I appreciate the effort you've put into it.

One of our goals for 1.0 and beyond is to eliminate as much global state and mutable state as possible.  Additionally, it provides only marginal benefits over simply injecting $location and asking for the .search().   Because of this, I can't accept this new `$stateSearch` service.  

I'd like to be able to support wildcard query parameters eventually, but I don't think this approach aligns with our current goals.

PR #2414, opened 2015-12-07
Update instruction to be consistent with examples.

RickCarlino:
There was a slight inconsistency in the `README`. The instructions say to add `ui-view` to the `<body>`, but the example shows `ui-view` on a wrapper div. This is a small tweak to keep things consistent.


nateabele:
This just seems unnecessarily wordy to me. `uiView` works as an element or an attribute, so the text is consistent with the example.

RickCarlino:
Sounds good.

PR #2435, opened 2015-12-29
feat(resolve): add $resolve service

wesleycho:
- Add `$resolve` service for exposing helper as an injectable

This is a WIP, and is in need of refinement. This also needs tests obviously, and perhaps documentation needs to be restored. Perhaps this also needs to be moved to a separate file.

Thoughts/comments?

### DIFF
--- a/src/ng1/angular1.ts
+++ b/src/ng1/angular1.ts
@@ -14,7 +14,12 @@
 import {IQService} from "angular";
 import {Router} from "../router";
 import {services} from "../common/coreservices";
-import {isObject} from "../common/common";
+import {forEach, isObject} from "../common/common";
+import {RawParams} from "../params/interface";
+import {Node} from "../path/module";
+import {Resolvables} from "../resolve/interface";
+import {Resolvable, ResolveContext} from "../resolve/module";
+import {State} from "../state/module";
 
 let app = angular.module("ui.router.angular1", []);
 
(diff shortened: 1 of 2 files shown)


src/ng1/angular1.ts
```diff
@@ -80,21 +83,41 @@ function ng1UIRouter($locationProvider) {
...
     bindFunctions(["replace", "url", "path", "search", "hash"], $location, services.location);
-    bindFunctions([ 'port', 'protocol', 'host'], $location, services.locationConfig);
+    bindFunctions(['port', 'protocol', 'host'], $location, services.locationConfig);
     bindFunctions(['baseHref'], $browser, services.locationConfig);
 
     return router;
   }
 }
 
+function resolveProvider() {
```
christopherthielen: This probably only needs to be a service, not a provider

src/ng1/angular1.ts
```diff
@@ -80,21 +83,41 @@ function ng1UIRouter($locationProvider) {
...
 angular.module('ui.router.init', []).provider("ng1UIRouter", <any> ng1UIRouter);
 // Register as a provider so it's available to other providers
 angular.module('ui.router.util').provider('$urlMatcherFactory', ['ng1UIRouterProvider', () => router.urlMatcherFactory]);
 angular.module('ui.router.router').provider('$urlRouter', ['ng1UIRouterProvider', () => router.urlRouterProvider]);
 angular.module('ui.router.state').provider('$state', ['ng1UIRouterProvider', () => router.stateProvider]);
+angular.module('ui.router.resolve', []).provider('$resolve', <any> resolveProvider);
 
 /* This effectively calls $get() to init when we enter runtime */
 angular.module('ui.router.init').run(['ng1UIRouter', function(ng1UIRouter) { }]);
+angular.module('ui.router.resolve').run(['$resolve', function(resolve) { }]);
```
christopherthielen: this shouldn't be necessary since there is no initialization in the $resolve service.

src/ng1/angular1.ts
```diff
@@ -80,21 +83,41 @@ function ng1UIRouter($locationProvider) {
...
     return router;
   }
 }
 
+function resolveProvider() {
+  this.$get = $get;
+  function $get() {
+    return {
+      resolve: (invocables, locals, parent, self) => {
+        let context = new ResolveContext([]);
```
christopherthielen: Give this a path with false `state`, i.e.:

```
let node = { state: {} }
let context = new ResolveContext([node]);
```

Then, you can add resolvables to that node like so:
```
    context.addResolvables(resolvables, node.state);
```

PR #2441, opened 2016-01-05
fix bug: cannot read property 'params' of undefined

ivchenkodima:


nateabele:
Can you post a plunkr or submit a unit test that triggers the issue? Also: http://prjs.radify.io/#/angular-ui/ui-router/pulls/2441 — Thanks!

PR #2447, opened 2016-01-07
fix(uiSrefActive): update the active classes when compiled

fpipita:
This PR makes sure the active classes are applied during the directive compilation phase so that they are available even after `$stateChangeSuccess` has already been fired.

This patch was originally developed for `1.0.0alpha0`, but since the test for it didn't pass on master either, I've backported it.

Plunkr showing the issue: https://plnkr.co/edit/z3kdm41eyuXK052fsflH?p=preview


christopherthielen:
Looks good, thanks

PR #2448, opened 2016-01-07
fix(uiSrefActive): update the active classes when compiled

fpipita:
This PR makes sure the active classes are applied during the directive compilation phase so that they are available even after `$stateChangeSuccess` has already been fired.

Plunker showing the issue: https://plnkr.co/edit/z3kdm41eyuXK052fsflH?p=preview

### DIFF
--- a/src/ng1/stateDirectives.ts
+++ b/src/ng1/stateDirectives.ts
@@ -329,6 +329,7 @@ function $StateRefActiveDirective($state, $stateParams, $interpolate) {
 
       function exactMatch(state, params) { return $state.is(state.name, params); }
 
+      update();
     }]
   };
 }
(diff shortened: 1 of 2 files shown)


christopherthielen:
Hi, thanks for the PR.    Is there a related issue ticket that it closes?

Your plunker seems to be working fine for me.  However, I see what your PR is fixing and it looks good to me at first glance.  Can you rebase against the current `feature-1.0` and squash commits?

fpipita:
Hi, I rebased the PR against `feature-1.0` but I'm not quite sure about what it should be squashed against, could you please explain me? :)

The plunker is not working because as shown in the image below, the text "I should get the active class:/" is not getting the active class.
plunkr

Issue #2392 is related to this one even though it referred the `master` branch.

christopherthielen:
> what it should be squashed against

my mistake, no squash necessary

> the text "I should get the active class:/" is not getting the active class.

aah, I see now.

Thanks!

PR #2464, opened 2016-01-15
fix(transitionTo): re-added the saved hash before broadcasting event

blokfyuh:
Re-added the saved hash before broadcasting $stateChangeStart. This way, libraries using this event to do their magic will have the hash accessible through toParams. (e.g. https://github.com/Narzerus/angular-permission)

Thoughts/comments ?

### DIFF
--- a/src/state.js
+++ b/src/state.js
@@ -1048,7 +1048,10 @@ function $StateProvider(   $urlRouterProvider,   $urlMatcherFactory) {
 
       // Filter parameters before we pass them to event handlers etc.
       toParams = filterByKeys(to.params.$$keys(), toParams || {});
-
+      
+      // Re-add the saved hash before we start returning things or broadcasting $stateChangeStart
+      if (hash) toParams['#'] = hash;
+      
       // Broadcast start event and cancel the transition if requested
       if (options.notify) {
         /**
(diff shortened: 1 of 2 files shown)


nateabele:
Yeah, the patch makes sense. We just need a matching unit test to prevent regressions.

blokfyuh:
Hi guys,

I'm not an expert regarding contributions, so what kind of unit tests would you expect for this patch ? I me

blokfyuh:
Hi guys,

I'm not an expert regarding contributions, so what kind of unit tests would you expect for this patch ? I mean wouldn't be redundant with the tests that @mismith wrote for #1867 ?

nateabele:
So, the idea is you want to have a unit test that fails against master but passes with your patch. In this case, that means asserting that the hash is available in a `$stateChangeStart` hook.

nateabele:
Test looks good. Almost there: http://prjs.radify.io/#/angular-ui/ui-router/pulls/2464

blokfyuh:
The subject looks good, the commits have been squashed. Do I pass ? :D

nateabele:
Almost! It's the subject of the commit, not the PR. Sorry about that. I'm gonna fix the language in the UI so it's clearer.

blokfyuh:
No problem. Let me know if we are good this time.

nateabele:
Yesssss!

Issue #2483, opened 2016-01-25
isOptional is always true


swisstofu:
When checking this line of code (https://github.com/angular-ui/ui-router/blob/02e98660a80dfd1ca4b113dd24ee304af91e9f8c/src/urlMatcherFactory.js#L923)

```javascript
if (type.name === "string" && !arrayMode && location === "path" 
  && config.value === undefined)
  config.value = ""; // for 0.2.x; in 0.3.0+ do not automatically default to ""
var isOptional = config.value !== undefined;
```

i noticed that if you have a definition like `url: '/{language:(?:fr|en|de)}'`, language will be treated as an optional parameter because of code above

i didn't find any workaround for now, but would it be possible instead to write 

```javascript
if (type.name === "string" && !arrayMode && location === "path" 
  && config.value === undefined)
  config.value = ""; // for 0.2.x; in 0.3.0+ do not automatically default to ""
var isOptional = !!config.value;
```

any advice ? 

thanks

swisstofu:
hello, a PR has been created https://github.com/angular-ui/ui-router/pull/2492 with a unit test regarding this bug.

christopherthielen:
This code is only in `legacy` where path parameters are optional by default.  We can't change that behavior until version 1.0.  In the 1.0 code, this should be working as expected, with the parameter being non-optional.

PR #2492, opened 2016-01-27
fix(UrlMatcher) isOptional

swisstofu:
fix for isOptional is always true #2483 and unit test

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -920,8 +920,6 @@ function $UrlMatcherFactory() {
     type = getType(config, type, location);
     var arrayMode = getArrayMode();
     type = arrayMode ? type.$asArray(arrayMode, location === "search") : type;
-    if (type.name === "string" && !arrayMode && location === "path" && config.value === undefined)
-      config.value = ""; // for 0.2.x; in 0.3.0+ do not automatically default to ""
     var isOptional = config.value !== undefined;
     var squash = getSquashPolicy(config, isOptional);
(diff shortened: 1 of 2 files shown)


nateabele:
Thanks for doing this patch, but please don't modify versioning or release files. Those are for the release process only. Also, the formatting on the test looks a bit off.

swisstofu:
sure sorry for that. everything should be fixed now.

christopherthielen:
Thanks for the PR.  

I don't think the conditional `value !== ""` is the right change.  
It seems like it would break default value of "" (which is very common; it's the default in 0.2.x)

I'm also surprised we didn't have a unit test for a default value of ""! 

```
url: '/foo/:str',
params: {
  str: { value: "" }
}
```
or to use an example with a regexp, which 

```
url: '/foo/{param:(?:foo|bar|)}',
params: {
  param: { value: "" }
}
```

swisstofu:
well i think then we can just get rid of the line

```javascript
if (type.name === "string" && !arrayMode && location === "path" && config.value === undefined)		
  config.value = ""
```
i've added some tests you suggested as well as mine and none are failing

christopherthielen:
http://prjs.radify.io/#/angular-ui/ui-router/pulls/2492

Can you squash your commits and revert the changes to the `release/*` files?

swisstofu:
done

christopherthielen:
I merged the tests into 1.0 (`master`); the deleted code from the PR was already deleted in the 1.0 codebase.

PR #2497, opened 2016-01-28
feat(ui-view): Add noanimation attribute to specify static renderer.

### LINKED ISSUE #2485: $stateChangeSuccess is fired before previous state $scope is destroyed in 0.2.16
This fiddle shows the difference in order of $destroy and $stateChangeSuccess between 0.2.15 and 0.2.16

https://jsfiddle.net/n3emkeo7/

This change broke our application, and I think it could be quite a significant breaking change.

My guess is that it's something to do with 1be13795 (unconfirmed).

Our use case:

We have a navigation control which listens for `$stateChangeSuccess` and subsequently updates the URL using `$state.go`.

In 0.2.15 this was fine if you navigated to a different page, the navigation control was destroyed before the `$stateChangeSuccess` event was fired, so the control would only redirect the user if they were navigating to a child state.

In 0.2.16 the `$stateChangeSuccess` event is fired before the control is destroyed, causing the user to be trapped on the page with the navigation control regardless of if they are trying to navigate away from the page.

christopherthielen:
fix(ui-view) use static renderer when animations are not enabled
closes #2485

### DIFF
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -1,3 +1,5 @@
+var ngMajorVer = angular.version.major;
+var ngMinorVer = angular.version.minor;
 /**
  * @ngdoc directive
  * @name ui.router.state.directive:ui-view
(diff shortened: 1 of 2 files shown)


src/viewDirective.js
```diff
@@ -133,24 +138,35 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate)
...
-      };
+    var statics = {
+      enter: function (element, target, cb) { target.after(element); cb(); },
+      leave: function (element, cb) { element.remove(); cb(); }
     };
 
+    if (!!attrs.noanimation) return statics;
+
+    function animEnabled(element) {
+      if (ngMajorVer === 1 && ngMinorVer >= 4) return !!$animate.enabled(element);
```
christopherthielen: disabled anim on a per-element basis

Issue #2501, opened 2016-01-30
Route clashing issue in chrome after 0.2.16


cdriscol:
I have a routing use case where I have nested/wizard states that must keep the same URL.  

For example:
* a user is taken to a create wizard with URL `/something/create`
* navigating between steps should not change the URL
* navigating directly to step 1 should be possible

I was able to do this by defining the following states (_Also, I'm very open to better solutions for this use case_):

```
$stateProvider
          .state('something.create', {
            url: '/create',
            abstract: true,
            template: '<div ui-view></div>'
          }) 
          .state('something.create.1', {
            url: '',
            template: '...'
          })
         // ...
          .state('something.create.4', {
            url: '',
            template: '...'
          });
```

This pattern has worked in versions 0.2.15 and lower.. but https://github.com/angular-ui/ui-router/pull/1585 added a `Array.prototype.sort` to the route rules.  The sort algorithm implemented in Chrome is not stable (https://bugs.chromium.org/p/v8/issues/detail?id=90).. so if you have a lot of routes defined in your application, then routes with the same prefix end up getting put into a somewhat random order.

I have added a link to 2 code pens which show this issue.  You can view them in debug mode to see the URLs.  You can notice the issue when you click the `hij.step.1` button after loading the pen.. it will take you to step 3 in Chrome.  Also, in debug mode if you add `#/hij` to the URL, it will take you to step 3 by default in Chrome.

v0.2.16: http://codepen.io/cdriscol/pen/adGBoX
v0.2.15: http://codepen.io/cdriscol/pen/bEKoXL

I think sorting rules by prefix length to avoid clashes is appropriate, but I think we should ensure a stable sort so that the first rule that matches is the first rule that is defined in the application.

cdriscol:
This also looks related to #2496

ArtworkAD:
:+1: just experience the same

christopherthielen:
See our discussion in PR #2502

PR #2502, opened 2016-01-30
Ensuring stable sort of route rules used for matching URLs.

cdriscol:
This is the fix for #2501.

### DIFF
--- a/src/urlRouter.js
+++ b/src/urlRouter.js
@@ -64,6 +64,7 @@ function $UrlRouterProvider(   $locationProvider,   $urlMatcherFactory) {
    */
   this.rule = function (rule) {
     if (!isFunction(rule)) throw new Error("'rule' must be a function");
+    rule.position = rules.length;
     rules.push(rule);
     return this;
   };
@@ -308,7 +309,8 @@ function $UrlRouterProvider(   $locationProvider,   $urlMatcherFactory) {
     rules.sort(function(ruleA, ruleB) {
       var aLength = ruleA.prefix ? ruleA.prefix.length : 0;
       var bLength = ruleB.prefix ? ruleB.prefix.length : 0;
-      return bLength - aLength;
+      var lengthDiff = bLength - aLength;
+      return lengthDiff !== 0 ? lengthDiff : ruleA.position - ruleB.position;
     });
 
     if (!interceptDeferred) listen();
--- a/test/urlRouterSpec.js
+++ b/test/urlRouterSpec.js
@@ -59,7 +59,31 @@ describe("UrlRouter", function () {
           match = ['/foo/bar', $match];
         }).when('/bar', function($match) {
           match = ['/bar', $match];
-        });
+        }).when('/', function() { 
+          match = '/home';
+        }).when('/abc', function() { 
+          match = '/abc/step1';
+        }).when('/abc', function() { 
+          match = '/abc/step2';
+        }).when('/defg', function() { 
+          match = '/defg';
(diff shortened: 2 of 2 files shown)


test/urlRouterSpec.js
```diff
@@ -59,7 +59,31 @@ describe("UrlRouter", function () {
           match = ['/foo/bar', $match];
         }).when('/bar', function($match) {
           match = ['/bar', $match];
-        });
+        }).when('/', function() {
```
cdriscol: These rule definitions may seem abundant - but it was the only way I could consistently reproduce the stable sorting issue in Chrome.

test/urlRouterSpec.js
```diff
@@ -84,6 +108,12 @@ describe("UrlRouter", function () {
       expect(match[0]).toBe("/foo/:param");
       expect(match[1]).toEqual({param: 'baz'});
     });
+    
+    it('should default to the first rule defined for same url', function() {
```
cdriscol: This test always passed in PhantomJs, but when run in Chrome (`grunt karma:debug`) this test would fail (before I implemented the stable sort fix).

nateabele:
Wow, nice! Thanks for doing the research on this. We were just gonna roll back #1585, but this seems like a better solution for sure. /cc @christopherthielen

src/urlRouter.js
```diff
@@ -308,7 +309,8 @@ function $UrlRouterProvider(   $locationProvider,   $urlMatcherFactory) {
     rules.sort(function(ruleA, ruleB) {
       var aLength = ruleA.prefix ? ruleA.prefix.length : 0;
       var bLength = ruleB.prefix ? ruleB.prefix.length : 0;
-      return bLength - aLength;
+      var lengthDiff = bLength - aLength;
+      return lengthDiff !== 0 ? lengthDiff : ruleA.position - ruleB.position;
```
cdriscol: Ensures a stable sort in Chrome browsers

christopherthielen:
I really do like the idea of ordering the rules by specificity _by default_.  We should definitely do that for 1.0.  

My main concern with #1585 is that it's a legitimate breaking change from 0.2.15.  For years we've told people that url rules are checked in the order that they were registered, and that they should re-order the registration of those rules if they need to change match priority.

Because sorting rules was a BC for 0.2.x, however, I think we should do one of three things for 0.2.18: 
- roll back #1585 completely
- allow users to select which strategy
  - insertion order
  - specificity sort order, including #2502
- allow users to provide their own sorting algorithm

nateabele:
Yeah, fair enough. @cdriscol Update the PR to point to `feature-1.0` and we'll merge it. @christopherthielen Yeah, fair enough. We'll just roll back #1585 for .18, merge this for 1.0, and implement sort strategies when we get time.

chino23:
could this fix be released with a quick minor update soon?

cdriscol:
@nateabele - I will get this into the `feature-1.0` branch when I get some time in the next couple days.

will be good to get the sort rolled back in 0.2.18

sebastiannm:
It would be good to release a new version for this soon. I found that it breaks a lot of stuff

nateabele:
@sebastiannm Yup, that's why it's going into 1.0 only, and the previous version is being rolled back.

JetFault:
@nateabele When do ya think the 0.2.18 release will come out? Just noticed this broke my app so had to revert to .15

nateabele:
@JetFault Hard to say. We have this and one or two other BC breaks to patch up.

PR #2507, opened 2016-02-01
2504: correctly formatted parent/child query parameters

pavel-blagodov:
I'm not sure that this is the most beautiful way of solving the issue but all tests are passed.


nateabele:
Thanks for digging into this. I'm not too keen on any change which increases the API surface area, but I'll tackle it myself later today or tomorrow.

Issue #1031, opened 2014-04-17
Refer to current state from ui-sref (e.g. for setting url parameters).


olegskl:
The [documentation](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.directive:ui-sref) explicitly states _"You can also use relative state paths within ui-sref, just like the relative paths passed to $state.go()."_ However, even though we are able to refer to the current state with `$state.go('.', {foo: 'bar'})` it doesn't seem to work when using `ui-sref` directive. I naively tried `ui-sref=".({foo:'bar'})"` which failed.

Of course it is possible to wire up `ng-click` and `$state.go`, but that leaves us with href-less anchors, which is not very useful from the user perspective.

I've [ventured a related question on stackoverflow](http://stackoverflow.com/questions/23106493/change-a-url-parameter-in-the-current-state-from-ui-sref-in-angular-ui-router) in case somebody had figured it out already. But it appears not to be the case.

Does this functionality exist already (maybe I just missed/misused it)? If not, should it be implemented?

albertpeiro:
+1 on this. Pretty pressing for handling user-friendly async pagination.

jaroslavrehorka:
+1

conorgriffin:
+1

demisx:
+1

bennewton999:
+1

zschiffelbein:
+1 using ui-sref="{foo: 'bar'}" renders the href tag on the first state and does not continue to update on state changes.

nateabele:
Okay, so sounds like maybe the `$watch()` is wrong. If somebody can post me up a quick Plunkr I'll take a look at it.

zschiffelbein:
Pretty sloppy, but it should give you an idea. The expected outcome is for it to stay on either the color or size route when you click on a different item. http://plnkr.co/edit/fARlus6UmokOsdRxbaoS

p3rsist:
I just updated the plunk with AngularJS 1.3.15 and ui-router 0.3.14 and the href tag still does not get updated on state change. http://plnkr.co/edit/8jZCbEVuoO4srwU8fIY5?p=preview

nateabele:
Okay, I see the issue. Should be able to patch it later today.

zivc:
@nateabele any luck?

darynmitchell:
Confirm I am able to switch between items in my current state by changing one param value using
```
<a ui-sref="{ itemId: myItem.id }">
```

[...]

PR #2541, opened 2016-02-10
Refer to current state from ui-sref (e.g. for setting url parameters). #1031

shprota:
Added minor modifications for current state reference via "." notation.


Maximaximum:
It would be useful if you could merge this pull request. I can't upgrade to ui-router 1.0 right now, but this fix would greatly help.

PR #2560, opened 2016-02-16
Updated API Reference documentation for ui-sref directive

webgene:
ui-sref is assigned to the state incorrectly, only parameter object is passed. Corrected by assigning the contacts.detail state with the required parameter.


christopherthielen:
The original is actually correct.   It demonstrates how to create a ui-sref that only changes parameter values.

PR #2570, opened 2016-02-19
feat(uiView): Fire the $onInit hook

### LINKED ISSUE #2559: $onInit hook from Angular 1.5.0 will not fire for a state controller
```
this.$onInit = function onInit() {
   // this will not fire when this is a state controller
};
```

See similar issue in [angular-ui-bootstrap #5472](https://github.com/angular-ui/bootstrap/issues/5472).

dgsmith2:
After instantiating the controller, will fire its $onInit if defined.

Closes #2559

### DIFF
--- a/src/ng1/viewDirective.ts
+++ b/src/ng1/viewDirective.ts
@@ -2,7 +2,7 @@ var ngMajorVer = angular.version.major;
 var ngMinorVer = angular.version.minor;
 /** @module view */ /** for typedoc */
 import {extend} from "../common/common";
-import {isDefined} from "../common/predicates";
+import {isDefined, isFunction} from "../common/predicates";
 import {trace} from "../common/trace";
 import {ViewConfig} from "../view/view";
 import {UIViewData} from "../view/interface";
@@ -319,6 +319,7 @@ function $ViewDirectiveFill (  $compile,   $controller,   $interpolate,   $injec
           let locals = data.$locals;
           let controllerInstance = $controller(controller, extend(locals, { $scope: scope, $element: $element }));
           if (controllerAs) scope[controllerAs] = controllerInstance;
+          if (isFunction(controllerInstance.$onInit)) controllerInstance.$onInit();
           $element.data('$ngControllerController', controllerInstance);
           $element.children().data('$ngControllerController', controllerInstance);
         }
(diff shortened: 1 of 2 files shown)


christopherthielen:
I think this looks good.

A couple observations:
- the [docs](https://code.angularjs.org/1.5.0/docs/api/ng/service/$compile#-controller-) suggest that `$onInit` is only for components, and only mentions them in the context of DDOs
- the [release notes](http://angularjs.blogspot.com/2016/02/angular-150-ennoblement-facilitation.html) explicitly state the hook is for components/directives
- [angular-route does not fire `$onInit`](http://plnkr.co/edit/g6TVMsekVOTnXDZm9PXc?p=preview) on its ng-view controllers
- I'd rather not re-implement the logic already [found in $compile](https://github.com/angular/angular.js/blob/v1.5.0/src/ng/compile.js#L2431-L2436)

That said, I don't see much of a downside.

I'll be curious if ng-view does something similar.

dgsmith2:
This is a bit contrived, but a downside of **NOT** doing it is you lose the ability to reuse a (component) controller that implements $onInit in all these extra scenarios and more.

I submitted a similar PR to address it with modals: angular-ui/bootstrap#5509.  Currently, there are no objections, but your observations may be worth noting over there as well.

christopherthielen:
>  lose the ability to reuse a (component) controller that implements $onInit

As I think about it, the whole point of `$onInit` it to allow the `bindToController` or `bindings` to be done after controller constructor is invoked, and then call `$onInit` in order to use the bound data.

I don't think  `$onInit` is useful without some sort of `bindings` (or `$ctrl.$resolve`) support.

dgsmith2:
Are you opposed to an initialization hook in general or one name "$onInit" due to the unmet implications of 1.5 component controllers?

I've found that having a hook at the very least simplifies testing.  In situations in which a controller needs to make a request when it is initialized one only needs to mock the service call when testing the hook call.

christopherthielen:
I'm in favor of the $onInit hook but don't think it's necessary until we support binding to component in some manner.

petebacondarwin:
It might be worth noting that the new component router has its own hooks. Notably it calls `$routerOnActivate` when the component has been instantiated and wired up by the `ng-outlet` (which is equivalent to your `ui-view`).

PR #1454, opened 2014-10-15
feat($StateRefDirective) add compatibility for angular/material material-button directive

brettstack:
I'm using angular/material and noticed that their `material-button` directive was not playing nicely with `ui-sref`. If `material-button` has an `href` attribute, it adds an inner `<a>` tag, but if it does not, it adss an inner `<button>` tag.

In order to get this to work with ui-router's `ui-sref`, the `$StateRefDirective` directive needs a higher priority than `material-button` and some functionality needs to be moved to the `compile` and `preLink` phases so that the `href` gets applied before `maerial-button` compiles.

PR #2586, opened 2016-02-26
RFC: feat(filters): implement asHref filter

gregoriokusowski:
A global filter to make url generation without ui-sref
possible with third-party libraries, since wrapping stuff
with links isn't pretty/semantically nice

Usage:

"stateName" | asHref --- Translates to $state.href("stateName")

---

An example, is the use of ui-router with angular-material,
where md-button reads a href attribute in order to create a
link tag, and people is using this work-around:

`<a ui-sref="home"><md-button>Home</md-button></a>`
as in http://stackoverflow.com/questions/32435358/ui-sref-and-md-button-how-it-works

### DIFF
--- a/src/ng1/stateFilters.ts
+++ b/src/ng1/stateFilters.ts
@@ -36,6 +36,25 @@ export function $IncludedByStateFilter($state) {
   return  includesFilter;
 }
 
+/**
+ * @ngdoc filter
+ * @name ui.router.state.filter:asHref
+ *
+ * @requires ui.router.state.$state
+ *
+ * @description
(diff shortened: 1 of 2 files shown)


gregoriokusowski:
Question: Is that name ok?

Problem: `$state.href` is generating two slashes after the `#`, as shown in the following spec:

```
asHref filter should return a relative child state FAILED
	Expected '#//b' to be '#/b'.
```

eddiemonge:
I think putting a higher priority on ui-sref would also solve this without having to do a new filter. Or a lower priority on md-button

gregoriokusowski:
@eddiemonge, we already have this higher priority approach (#1454). I think we can close this PR if that gets merged. WDYT?

PR #2600, opened 2016-03-02
fix(transition/transitionService): uses console.error to log error in default error handler

matthewjh:
Hey,

We should use `console.error` here rather than `console.log` to benefit from the browser's error-specific logging features (e.g. stack trace display).


nateabele:
:+1:

PR #2602, opened 2016-03-02
fix(src/resolve): use injector's strictDi value in calls to .annotate

matthewjh:
This PR passes through `$injector.strictDi` to `$injector.annotate` calls. At present, if a resolve function is unannotated and `$injector.strictDi` is `true`, an error is not thrown in contrast to the expected behavior for angular injectables.


matthewjh:
The tests pass locally but fail on Travis. I'll investigate.

matthewjh:
Fixed it. :)

@nateabele, mind taking a look?

PR #2624, opened 2016-03-15
docs(Transition): update $state local injectable

eddiemonge:
`$state`, not `$state$` for hook injectables

### DIFF
--- a/src/transition/interface.ts
+++ b/src/transition/interface.ts
@@ -372,7 +372,7 @@ export interface IHookRegistry {
    * @example
    * ```
    *
-   * $transitions.onEnter({ to: 'admin' }, function(AuditService, $state$, $transition$) {
+   * $transitions.onEnter({ to: 'admin' }, function(AuditService, $state, $transition$) {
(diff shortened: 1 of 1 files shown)


christopherthielen:
`$state$` is correct, but it's missing in the docs for `onEnter` (and likely `onRetain` and `onExit` too)

It should be near 
>  * The current [[Transition]] can be injected as `$transition$`.
and should say something like
The [[State]] being entered can be injected as `$state$`

eddiemonge:
was it recently added because it wasn't available in the actual alpha release

christopherthielen:
It should have been in the alpha release.  Let me check if there are any unit tests

christopherthielen:
https://github.com/angular-ui/ui-router/blame/97765440595d666b1490d8b1db70795e406f740e/test/transitionSpec.ts#L152-L175

That was definitely in the alpha

eddiemonge:
Oh maybe its because I was trying to use onFinish and not onEnter

PR #2636, opened 2016-03-21
[WIP] fix: $stateChangeStart with event.preventDefault()

masterspambot:
In order to properly process preventing event handling in libs like angular-permission earlier breaking of transitionTo is required to avoid doubled state resolving and allow async calls combined with access to states.


nateabele:
Yeah, sorry. State events are deprecated in 1.0, so we didn't want to introduce any additional things that would just be going away...

nateabele:
I haven't used either of those libraries so I can't speak to that. @christopherthielen might know.

PR #2642, opened 2016-03-23
Konetix

Konetix:
That resolution can solve issue connected with not working transition animation


nateabele:
Sorry, I don't understand what this is about. Please follow the contributor guide and open an issue with a reproducible test case. Thanks.

PR #2681, opened 2016-04-08
your package name is wrong in package.json

mkoryak:


christopherthielen:
hi @mkoryak 

this repository is now the  repository for `angular-ui-router`, `ui-router-core`, and `ui-router-ng2`.  I renamed this package.json to `ui-router` so it would not be mistaken for the angular 1 package.json, which is in `/packages/ng1/package.json`

mkoryak:
but you should pick a name that you own, no? 
see `pad-left`

PR #2706, opened 2016-04-20
Update README.md

Leinad93160:


ocombe:
I don't think it's a typo :) "thus" is the right word here

PR #2721, opened 2016-04-29
fix(uiSref, uiState): added click unbind to prevent memory leaks

smelukov:
Fixed memory leaks beetween states with many ui-sref in their templates


smelukov:
In my case, i have many links with ui-sref attribute.
For every transfer beetween states was allocated 180kb of memory that does not freed.
After that fix I've got 17kb instead of 180kb

nateabele:
Seems 👍 . Can anybody else confirm before we move on?

smelukov:
no one obviously...

christopherthielen:
Seems like it *shouldn't* be necessary, but I have no issue with merging it

PR #2727, opened 2016-05-04
refactor(uiView): just promisify animation and simplify some code

smelukov:
`$q.defer()` is used for capability, cause older versions of angular is not supported `$q(function(resolve) {...})` and `$q.resolve()`

### DIFF
--- a/src/viewDirective.js
+++ b/src/viewDirective.js
@@ -154,25 +154,45 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate,
   function getRenderer(attrs, scope) {
     var statics = function() {
       return {
-        enter: function (element, target, cb) { target.after(element); cb(); },
-        leave: function (element, cb) { element.remove(); cb(); }
+        enter: function (element, target) {
+          target.after(element);
+
+          return $q.when();
+        },
+        leave: function (element) {
+          element.remove();
+
+          return $q.when();
+        }
       };
     };
 
     if ($animate) {
       return {
-        enter: function(element, target, cb) {
+        enter: function(element, target) {
           if (angular.version.minor > 2) {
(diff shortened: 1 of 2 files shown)


Wykks:
Not sure about your usage of `defer();`

```
      enter: function (element, target) {
          var defer = $q.defer();

         target.after(element);
         defer.resolve();

          return defer.promise;
        },
```
Should be written like this: 

```
      enter: function (element, target) {
         target.after(element);
         return $q.resolve();
        },
```

smelukov:
$q.defer() is used for capability, cause older versions of angular is not
supported $q(function(resolve) {...}) and $q.resolve()

2016-05-12 17:17 GMT+03:00 Wykks <notifications@github.com>:

Wykks:
Sorry, deleted my comment by mistake !
Indeed `$q.resolve();` is not supported, but `$q.when();` could be used instead (`$q.resolve();` is just an alias to `$q.when();`)

smelukov:
You're right! I wasn't thought about that.
Fixed ;)

src/viewDirective.js
```diff
@@ -209,29 +237,6 @@ function $ViewDirective(   $state,   $injector,   $uiViewScroll,   $interpolate,
 
         updateView(true);
 
-        function cleanupLastView() {
-          if (previousEl) {
-            previousEl.remove();
```
christopherthielen: This code is important, and I don't see it in the new impl.  It ensures that the prior ui-view copies get properly destroyed.

This code allows views to be swapped in multiple times, even while the first instance is still animating out.  This can be seen by creating a very long "out" animation, then cycling back and forth between sibling states.

christopherthielen:
overall I like your refactor.  See my diff comment about some cleanup I think this refactor has removed.

Also, make sure to also target the `master` branch

PR #2790, opened 2016-06-03
ng2 - Enable change detection for route components

isaacplmann:
Newly created components need an initial change detection to be fired manually.

Perhaps this is only a problem when the components are using the OnPush ChangeDetectionStrategy.  Without this fix, components don't load data until some browser event fires - i.e. click, etc.


christopherthielen:
thanks!

PR #2821, opened 2016-06-20
feat($state.fromUrl): New function in $stateProvider

xaviml:
A method to resolve a list of states and its parameters from a given URL. This is an example:

```
var app angular.module('app', ['ui.router']);
app.controller('ctrl', function ($scope, $state, somePathFromServer) {
  $scope.go = function(){
    var target = $state.fromUrl(somePathFromServer);
    if(target.length == 0)
      $state.go(target[0].name, target[0].args);
  }
});
```


clementprevot:
+1 for this feature

PR #2834, opened 2016-06-24
feat(urlMatcherFactory): add type "urlEncode" param

clementprevot:
This parameter allow to ignore the URL encoding of the URL parameters
when generating an href.
The use case for this is to allow to have a state param that contains
"/" that should be restored as real "/" in the generated URL path.

### DIFF
--- a/src/urlMatcherFactory.js
+++ b/src/urlMatcherFactory.js
@@ -342,8 +342,8 @@ UrlMatcher.prototype.format = function (values) {
 
   var i, search = false, nPath = segments.length - 1, nTotal = params.length, result = segments[0];
 
-  function encodeDashes(str) { // Replace dashes with encoded "\-"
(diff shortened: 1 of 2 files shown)


cvn:
I'm not on the ui-router team, but if anyone else is looking for this feature, it's available in 1.0 with "raw" types. e.g.

```js
.config(function ($urlMatcherFactoryProvider, $stateProvider) {
    $urlMatcherFactoryProvider.type('myType', {
        raw: true // do not encode
    });
    $stateProvider.state('myState', {
        url: '/{foo:myType}'
    });
})
```

See #2218

christopherthielen:
If somebody wants to backport the `raw` api from 1.0, I'll merge it

felixfbecker:
@clementprevot are you planning to backport `raw` to 0.x?

clementprevot:
Hum, no I'm not, sorry :wink:

PR #2835, opened 2016-06-24
fix(history): back button with preventDefault in $stateChangeStart causes history to be overwritten #1525

olibaron:
preventDefault() in the event $stateChangeStart causes history to be overwritten when the back button is clicked #1525
Fixed by changing the update routine for 'defaultPrevented' updates. The update trigger now uses a string instead of a bool argument to distinguish 'read', 'defaultPrevented' and normal update events. 
When using the back button is this scenario the $location.url(location) method is omitted so the history does not get overwritten anymore. The orginal scenarios ar not effected by this fix.


olibaron:
The file  release/angular-ui-router.js contains changes generated by grunt. As this is a generated file, these changes can of course be ommited. the src files refelect the changes and seem to be working with every test I did. The previous and next buttons seem to be working very well now. I hope these changes get accepted.

olibaron:
When using prteventDefault() in $stateChangeStart, a back navigation caused an update trigger which caused the history to be overwritten by $location.url(location). The trigger already knew that $stateChangeStart was defaultPrevented but the update trigger did not distinguish this event from the normal update where 'read' was false. 
I changed the boolean in the update trigger to a string argument so the 'defaultPrevented' update trigger can be distinguished from a 'read' or normal trigger. All seems to be working as should.
I updated the documentation of the source

nateabele:
I'm not seeing a test that fails against the existing code to actually prove the issue.

nateabele:
Also, this PR is pretty messy, even aside from the changes to build files: http://prjs.radify.io/#/angular-ui/ui-router/pulls/2835

olibaron:
Ok, I will add the test which will prove the existing code is failing. Also I will cleanup this pr. 
The issue is very clear. When using event.preventDefault in $stateChangeStart, the history gets overwritten when using the back button. So only back navigation is working and forward navigation gets disabled because of this.
This fix, fixes this and back and forward navigation will work as intended

PR #2841, opened 2016-06-28
window is actually unused

graingert:


graingert:
and you should use $window where you need it

graingert:
@christopherthielen bump

christopherthielen:
Thank you for the PR.  I'm not going to merge this, because we're no longer using the Gruntfile to do release builds.  (I'd like to eliminate grunt completely)

The current build uses webpack and is found in `./scripts/package.sh`

graingert:
I'm looking for this to be merged into the 0.2 branch which does use gulp

graingert:
I'm looking for this to support ui-router 0.2 which doesn't have
./scripts/package.sh

On 26 Jul 2016 8:16 pm, "Chris Thielen" <notifications@github.com> wrote:

Thank you for the PR. I'm not going to merge this, because we're no longer
using the Gruntfile to do release builds. (I'd like to eliminate grunt
completely)

The current build uses webpack and is found in ./scripts/package.sh

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/angular-ui/ui-router/pull/2841#issuecomment-235374791>,
or mute the thread

christopherthielen:
@graingert oh, legacy branch, I see that now.

What problem this PR solves for you?  How will `angular` differ from `window.angular`?  

This code has been [unchanged since the very first release 0.0.1](https://github.com/angular-ui/ui-router/blob/0.0.1/Gruntfile.js#L29), so I'm not convinced there's a problem with the way it currently is.  Please explain.

graingert:
This makes it easier to get ui-router to work under node.js and web workers.

PR #2886, opened 2016-07-22
Documentation link update

stevermeister:
each time I have to fix it in url manually :)


christopherthielen:
Thanks for the PR, but we're moving docs over to https://ui-router.github.io/docs/latest/

I updated the README.md

PR #2894, opened 2016-07-28
fix(StateService): remove jQuery deprecated feature

### LINKED ISSUE #2806: Using `bind()` method which is now deprecated in jQuery
The `bind()` and `unbind()` methods which `angular.element` implements from the jQuery API are now deprecated in jQuery. In the new [slim build of jQuery 3.0](https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/) these methods are removed from the code for the first time.

This means that if you use the new version of jQuery along with Angular and UI Router, Angular will [gracefully let jQuery take over from `angular.element`](https://docs.angularjs.org/api/ng/function/angular.element) but `bind()` and `unbind()` won't be available when UI Router tries to use them for the `$StateRefDirective` and `$StateRefDynamicDirective` link functions, and functionality will break. See this plunkr (and watch the console for errors):
http://plnkr.co/edit/b3pOQKRmuJRS7RyLbDP8?p=preview

(When you select a different state, it won't intercept the route change and a request will go off to the server, wrongly, because it wasn't able to set the event handler up.)

Replacing the (only 2!) usages of `bind()` and `unbind()` with `on()` and `off()` respectively should fix compatibility with the newest jQuery whilst maintaining compatibility with `angular.element`.

More than happy to do a PR for this, but will just wait for the nod from a project member.

vvakame:
fixes #2806

https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryfnbind-is-deprecated

### DIFF
--- a/src/ng1/directives/stateDirectives.ts
+++ b/src/ng1/directives/stateDirectives.ts
@@ -159,7 +159,7 @@ function $StateRefDirective($state, $timeout) {
       update();
 
       if (!type.clickable) return;
-      element.bind("click", clickHook(element, $state, $timeout, type, function() { return def; }));
+      element.on("click", clickHook(element, $state, $timeout, type, function() { return def; }));
     }
   };
 }];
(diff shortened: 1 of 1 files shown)


christopherthielen:
thanks

PR #2906, opened 2016-08-03
Legacy

flyonok:


christopherthielen:
I think this was trying to catch `master` up with commits in `legacy`, right? 

I've done this manually since these two branches have diverged.

PR #2990, opened 2016-09-13
fix(common): Remove `url()` from LocationService interface

warrenseymour:
- `LocationService.url()` has been deprecated in favour of `setUrl()`,
  but this interface had not been fully updated.


christopherthielen:
Huh, I thought `.url()` was still used as the read method, but apparently we're just using `.path()`

PR #2996, opened 2016-09-14
Replaced deprecated jquery functions

Dona278:
Since jQuery 3.0, both .bind() and .unbind() have been deprecated.
The suggested replacements are:
jquery.bind() => jquery.on()
jquery.unbind() => jquery.off()


christopherthielen:
Related #2894

PR #3018, opened 2016-09-19
fix(urlMatcherFactory): fix tilde edge case with "string" encoding

cvn:
Hello,

I noticed that "~2F" in a param value would be encoded as "~~2F" and then decoded as "~/" ([plunk](http://plnkr.co/edit/Nu2WgFTL36dQnetRByA6?p=preview)). I fixed this so now the encoding/decoding are symmetrical. I think this use case was the reason for encoding "~", and the decoding issue was an oversight.

Note: The problem was in the decode function (valFromString), but I also refactored the encode function (valToString) so that the functions stay easy to read as a pair.

Let me know what you think. If you like it, I also prepared a [branch](https://github.com/angular-ui/ui-router/compare/legacy...cvn:fix-tilde-encoding-legacy) that ports the changes to legacy.


christopherthielen:
a perfect PR, thank you!

PR #3028, opened 2016-09-22
ui-router api doc  

shashouzhiwang:
ui-router api doc


christopherthielen:
Hi, thanks for the PR!

Our web site is moved to: http://ui-router.github.io/ and the docs are moved to:
https://ui-router.github.io/docs/latest 
https://ui-router.github.io/docs/0.3.1

Because of that I can't accept this PR.  
In the future we will be removing these gh-pages docs adding redirects to the new site, but I don't have the time to address that right now.

Issue #3027, opened 2016-09-21
Bugfix in the legacy codebase


zeusdeux:
Would y'all be ok with a PR that is based on `legacy` branch?

The PR will address an issue with `ui-sref` where it incorrectly discards the promise returned by  `$state.go` and doesn't handle promise rejection.

It will also fix the erroneous behaviour outlined in #2123 if the state transition was triggered by `ui-sref` directive. Also, the change will be isolated to `stateDirective.js` so it shouldn't impede current dev efforts in any way.

Thanks!

christopherthielen:
yep! Is the issue fixed in 1.0 already?  If not, I'd like a PR for both `master` and `legacy`

zeusdeux:
[From the looks of it](https://github.com/angular-ui/ui-router/blob/master/src/ng1/directives/stateDirectives.ts#L67) it isn't fixed on master yet. 

I am unsure that my solution would apply to 1.0 though since it involves triggering a state change event. More specifically `$stateChangeCancel` when a transition is superseded.

christopherthielen:
> More specifically $stateChangeCancel when a transition is superseded.

Interesting... although it doesn't seem like that should be handled by the ui-sref directive.  It seems like the `transitionTo` should be doing that.  

Can you explain fully what the issue you're solving is?

zeusdeux:
I would absolutely love it if `transitionTo` broadcasted `$stateChangeCancel` on `$rootScope` when it rejects with `TransitionSuperseded`. The reasons why I chose not to do it in `transitionTo` are:

1. `transitionTo` method is quite involved and it is quite easy to introduce a bug in it
2. it would change `$stateChangeCancel` trigger semantics across the board
3. I wanted to keep the approach more silo-ed so that it didn't impact part of the codebase I don't know

Now, the issue I am trying to solve is quite straightforward. I am in some state `A` and I start a transition to state `B`. Before the transition to `B` completes, I transition back to `A`. In this scenario, in the legacy world, I am not notified of anything _and_ the locals for state `B` are resolved anyway. None of the `$stateXYZ` nor `$viewContentXYZ` events fire. As a consumer of the api, if I want to do something as simple as show a loader between transitions, the current behaviour would lead to the loader being stuck since no event is fired to notify me that the transition was cancelled.

zeusdeux:
@christopherthielen Any thoughts?

Thanks!

christopherthielen:
[...]

PR #3029, opened 2016-09-22
fix(ui-sref): handle state transition promise rejection

zeusdeux:
In the current world, `ui-sref` does nothing when the promise returned by
$state.go is rejected. This PR gives `ui-sref` directive the ability to
emit `$stateChangeCancel` when the transition promise rejects.
The reason why I have chosen to implement this in the `ui-sref` directive
and not in the definition for $state.go or $state.transitionTo are
outlined in the issue below.

#3027

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -26,14 +26,23 @@ function getTypeInfo(el) {
   };
 }
 
-function clickHook(el, $state, $timeout, type, current) {
+function clickHook(scope, el, $state, $timeout, type, current) {
   return function(e) {
     var button = e.which || e.button, target = current();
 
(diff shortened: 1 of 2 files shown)


christopherthielen:
@zeusdeux as mentioned in the issue #3027 I think this belongs in the `transitionTo` code.  If it only is implemented for ui-sref, the parity between stateChangeStart and (Success/Cancel/Error) is still broken.  

I'd be happy to help ensure the equivalent code in `transitionTo` is correct.

zeusdeux:
@christopherthielen Agreed. Working on a patch for `transitionTo`

zeusdeux:
@christopherthielen Done here #3039

Closing this in favour of that.

PR #3036, opened 2016-09-25
fix(stateDirective): using on to attach an event handler instead of bind

### LINKED ISSUE #3035: element.bind is not a function with jquery.slim version
The ```.bind()``` method has been deprecaten in [version 3.0+](http://api.jquery.com/bind/).

But in the **slim build** of jQuery the method has been removed directly. So, the uiSref directive fails when tries to attach a handler to the click event.

Plunker with the error in the console: https://plnkr.co/edit/2dceELuWmG2wq3027gwD

This only occurs with the slim build. This is not failing with jqLite or jQuery full build.

Can we add support to the slim build? jqLite also supports the ```.on()```` method. We just need to replace the bind with the on and this issue is resolved.

navarroaxel:
fix #3035

### DIFF
--- a/src/stateDirectives.js
+++ b/src/stateDirectives.js
@@ -145,9 +145,9 @@ function $StateRefDirective($state, $timeout) {
 
       if (!type.clickable) return;
       hookFn = clickHook(element, $state, $timeout, type, function() { return def; });
-      element.bind("click", hookFn);
+      element[element.on ? 'on' : 'bind']("click", hookFn);
       scope.$on('$destroy', function() {
-        element.unbind("click", hookFn);
+        element[element.off ? 'off' : 'unbind']("click", hookFn);
       });
     }
   };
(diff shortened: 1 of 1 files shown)


nateabele:
This is a BC-break. We can add support for `on` and `off`, but not by dropping support for `bind`/`unbind`.

navarroaxel:
Yeah, I can see it's a breaker... How do you like the IF?
````javascript
element.bind ? element.bind("click", hookFn) : element.on("click", hookFn);
````
or this:
````javascript
element.[element.bind ? 'bind' : 'on']("click", hookFn);
````
Should I try to use ```on``` before ```bind``` or viceversa?

christopherthielen:
I think I merged a similar commit into master

christopherthielen:
@navarroaxel we would want to try `on` before `bind`

navarroaxel:
Hey @christopherthielen, are you saying this commit is useless because it was covered by another one?

navarroaxel:
Hey @nateabele, I have change the code to be keep the BC.

christopherthielen:
I think I merged a similar commit on `master`, not `legacy`

christopherthielen:
Thanks!

nateabele:
Yup, thanks @navarroaxel.

PR #3056, opened 2016-10-01
fix(Ng1ViewDeclaration): Make controllerProvider IInjectable

### LINKED ISSUE #3044: controllerProvider on Ng1ViewDeclaration should be of type IInjectable instead of Function
I would expect this state definition to still be valid:

    .state("product.store.live", {
        url: "/live"
        views: {
            "storeView@product": {
                controllerProvider: ["options", storeControllerProvider], // <--- compilation fail
                templateUrl: "app/components/product/detail/shared/StoreView.html"
            }
        }
    })

Looking at the code in `ng1/statebuilders/views.ts#getController` the code looks for an ng1 array-notation-style array/function.

ciddan:
Closes #3044. Changes the type of controllerProvider in Ng1ViewDeclaration to IInjectable. Adds 1 test for controller provider in ng1/stateSpec. Updates the controller handling test in ng1/viewSpec to verify that controllerProvider is IInjectable .

### DIFF
--- a/src/ng1/interface.ts
+++ b/src/ng1/interface.ts
@@ -360,7 +360,7 @@ export interface Ng1ViewDeclaration extends _ViewDeclaration {
    * }
    * ```
    */
-  controllerProvider?: Function;
+  controllerProvider?: IInjectable;
 
   /**
    * The scope variable name to use for resolve data.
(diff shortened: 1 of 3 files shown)


christopherthielen:
very nice PR, thanks!

PR #3063, opened 2016-10-05
docs(Ng2StateDeclaration): fix views examples

Awk34:
views properties should be of the format `{component: MyComponent}`

I believe this:

```js
views: {
  header: HeaderComponent,
  body: BodyComponent,
  footer: FooterComponent
}
```
should be this:
```js
views: {
  header: {component: HeaderComponent},
  body: {component: BodyComponent},
  footer: {component: FooterComponent}
}
```


christopherthielen:
Thanks!  You are correct, the `component: ` is currently necessary.  However, I'd like to also allow the shorthand version in the future.

PR #3067, opened 2016-10-05
fix(Resovable): Fixes #3066 (RXWAIT)

### LINKED ISSUE #3066: [Resolve] async ResolvePolicy "RXWAIT" does not behave as expected
When returning an errored observable from a resolve factory the state transition never happens (state home.observable-that-throws). Furthermore, the items that are emitted after the first item don't seem to be passed on to the stream that is resolved (home.observable).

Plunker: http://plnkr.co/edit/x2mTrhH3RQuNV4fcXSrO?p=preview
Expected: http://plnkr.co/edit/MXMAAWLDD6jGUz99SDlG?p=preview

Kamshak:
This fixes the issue, which i believe is caused by issues with the cache operator (https://github.com/ReactiveX/rxjs/issues/2006, https://github.com/ReactiveX/rxjs/issues/1959)I couldn't wrap my head around how to unit test this - help would be much appreciated. I have done some testing manually in a plunker (http://plnkr.co/edit/7T1afONEHI4qxHASyOn0).

### DIFF
--- a/src/resolve/resolvable.ts
+++ b/src/resolve/resolvable.ts
@@ -120,13 +120,15 @@ export class Resolvable implements ResolvableLiteral {
      * For RXWAIT policy:
      *
      * Given an observable returned from a resolve function:
-     * - enables .cache() mode (this allows multicast subscribers)
-     * - then calls toPromise() (this triggers subscribe() and thus fetches)
-     * - Waits for the promise, then return the cached observable (not the first emitted value).
+     * - waits for the first item to be emitted by using take
+     * - calls toPromise to make the stream eager
+     * - in the promise: creates a multicast behaviour stream that emits the first item and all subsequent item
+     * - enables refCount to make sure subscriptions are cleaned up correctly and connected to automatically
      */
     const waitForRx = (observable$: any) => {
-      let cached = observable$.cache(1);
-      return cached.take(1).toPromise().then(() => cached);
+      return observable$.take(1).toPromise().then((firstItem) => {
+      	return observable$.publishBehavior(firstItem).refCount();
+      });
     };
 
     // If the resolve policy is RXWAIT, wait for the observable to emit something. otherwise pass through.


christopherthielen:
Thanks for the PR!

I was chatting about this on [gitter](https://gitter.im/angular-ui/ui-router).  @adharris found that it still isn't quite right:   See this modified plunker: http://plnkr.co/edit/9R1m2cbrHVfOdvwsS9sE?p=preview

The resolve has been augmented by logging on `.do()`.  This code should only print `do 0` once when the `home.observable` state is activated but it prints twice.

The `RXWAIT` code definitely needs to be fixed.  It's turning out to be more challenging to get right than I originally thought.

adharris:
The duplicate log happens because there is at least one tick where there are 0 subscribers to the observable.  This causes the interval to stop ticking, and when it subscribed to in the controller it restarts from 0. If instead of logging it was an http call, it would result in duplicate calls.

I think to get the right behavior, RXWAIT will need to subscribe to the observable, and maintain that subscription for some period of time after the first value is emitted; but I haven't quite figured out the right time to release the subscription. If its released before the controller can subscribe, then it will cause duplicate requests.  I've had some success by waiting for the transition to finish and then waiting for the ngZone to stabilize; but that's not without issues either.

Kamshak:
Oh interesting, hadn't noticed that! Seems quite tricky indeed... Wouldn't it be possible to unsubscribe once the state has changed? This would mean that a subscription is always active as long as a state is active. I think this is fine - if for some reason this causes problems it would always be possible to not use RXWAIT and implement the logic in the Component/Controller directly. Thoughts?

adharris:
Ya, I think maintaining the subscription while the state is active is the same conclusion we came to.  I've been faking it like this:

```
export function fakeRXWait<T>(
        token: any,
        stream: Observable<T>,
        transition: Transition,
        transitions: TransitionService): Promise<Observable<T>> {

    return new Promise<Observable<T>>((resolve, reject) => {
        let connectable = stream.publishReplay(1);

        connectable.take(1).subscribe(
             (first) => resolve(connectable),
             (error) => reject(error));

        let subscription = connectable.connect();
        let stateName = getState(token, transition).name;
        let transitionUnsub = transitions.onSuccess({exiting: stateName}, () => {
            subscription.unsubscribe();
            transitionUnsub();
        });
    });
}
```

Which wraps an observable in a promise that resolves once the first element gets emitted, and then use the normal `WAIT` policy.

The trick is knowing when to unsubscribe, which I do with a one-off transition hook; but that requires the  state name to be known to the resolvable.  I fake it with a `getState` function, but my implementation of that is pretty specific to my app.

I think we need to add some way for a resolvable to clean itself up after the state exits.  Not sure what that looks like yet.

PR #3068, opened 2016-10-05
fix($InjectorLike) Optional generic type support

csvn:
When working with Typescript, you currently need to do something like this to set type correctly:

```typescript
$urlRouterProvider.otherwise(($injector) => {
  let myService: MyService = $injector.get('myService');
  myService.doSomething();
});
```

With this change, you could instead write the above like this:

```typescript
$urlRouterProvider.otherwise(($injector) => {
  $injector
    .get<MyService>('myService')
    .doSomething();
});
```

### DIFF
--- a/src/common/interface.ts
+++ b/src/common/interface.ts
@@ -36,6 +36,7 @@ export interface UIInjector {
    * @return the Dependency Injection value that matches the token
    */
   get(token: any): any;
+  get<T>(token: any): T;
 
   /**
    * Asynchronously gets a value from the injector
@@ -58,6 +59,7 @@ export interface UIInjector {
    * @return a Promise for the Dependency Injection value that matches the token
    */
   getAsync(token: any): Promise<any>;
+  getAsync<T>(token: any): Promise<T>;
 
   /**
    * Gets a value from the native injector
@@ -73,4 +75,5 @@ export interface UIInjector {
    * @return the Dependency Injection value that matches the token
    */
   getNative(token: any): any;
+  getNative<T>(token: any): T;
 }
--- a/src/common/coreservices.ts
+++ b/src/common/coreservices.ts
@@ -41,6 +41,7 @@ export interface $QLike {
 
 export interface $InjectorLike {
   get(token: any): any;
+  get<T>(token: any): T;
   has(token: any): boolean;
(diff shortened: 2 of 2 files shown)


christopherthielen:
nice! 
Can it be opt-in, though, or do you *have* to supply a type?

csvn:
It is opt in. The interface supports both versions:

```typescript
interface $InjectorLike {
  get(token: any): any;
  get<T>(token: any): T;
  // ...
}
```

As long as you match any of the above, Typescript should not complain.

csvn:
[See this simple Typescript playground example](https://www.typescriptlang.org/play/index.html#src=interface%20Foo%20%7B%0D%0A%09bar()%3A%20any%3B%0D%0A%09bar%3CT%3E()%3A%20T%3B%0D%0A%7D%0D%0A%0D%0Alet%20foo%3A%20Foo%20%3D%20%7B%0D%0A%09bar()%20%7B%0D%0A%09%09return%20null%3B%0D%0A%09%7D%0D%0A%7D%0D%0A%0D%0Alet%20any%20%3D%20foo.bar()%3B%0D%0Alet%20string%20%3D%20foo.bar%3Cstring%3E()%3B%0D%0A). If you hover over `any` it will be type `any`, whereas `string` will be type `string`.

christopherthielen:
super!  Would you mind adding it to the https://github.com/angular-ui/ui-router/blob/master/src/common/interface.ts (i.e., `transition.injector()`)

csvn:
Sure, no problem. I didn't notice that one.

csvn:
Did the same change to https://github.com/angular-ui/ui-router/blob/master/src/common/interface.ts, and squashed the commit.

src/common/interface.ts
```diff
@@ -36,6 +36,7 @@ export interface UIInjector {
    * @return the Dependency Injection value that matches the token
    */
   get(token: any): any;
+  get<T>(token: any): T;
 
   /**
    * Asynchronously gets a value from the injector
```
csvn: Np. That's what I get for not looking too hard :P

Fixed.

christopherthielen:
Thanks a lot!

JounQin:
How can I use this feature with `UrlService` due to deprecation of `UrlRouterProvider`?

PR #3155, opened 2016-11-16
use $templateRequest from AngularJS in ui-router 0.3.2

anantanandgupta:
- modified `$templateFactory` to use `$templateRequest` provider of
AngularJs in case it is avalable
- removed the test case which enforces the setting of request header
`Accept: text/html`


anantanandgupta:
@christopherthielen i am not sure why that build is failing I have run all the supporting version from 1.2.x

anantanandgupta:
ok ... good to go ... it was failing in Angular 1.0.8 ... fixed and tests passed.

christopherthielen:
Closing in favor of #3182

PR #3164, opened 2016-11-21
fix(uiView): do not leave initial view scope undestroyed

### LINKED ISSUE #1896: Default ui-view content with directive using required causes error
When using default ui-view content along with directives that use `require` the directives can't find the required controller and cause an error similar to the following.

https://docs.angularjs.org/error/$compile/ctreq?p0=mdTabs&p1=mdLabelTemplate

This can be seen in the latest version of angular-material if your using tabs but I'm sure many other libraries and custom directives make use of `require`.

jpekkala:
Previously the contents of the initial view were linked and left
undestroyed when the view was replaced with a subview. Because the view
was not destroyed, directives like ngInclude assumed it was safe to
compile and link asynchronous content to it. This would cause a ctreq
error if the asynchronous content required another directive from the
DOM. Now the initial view is either not linked at all or its scope is
properly destroyed.

Closes #1896

### DIFF
--- a/test/viewDirectiveSpec.js
+++ b/test/viewDirectiveSpec.js
@@ -88,7 +88,7 @@ describe('uiView', function () {
     controller: function() {
       this.someProperty = "value"
     },
-    template: "hi",
+    template: "{{vm.someProperty}}",
     controllerAs: "vm"
   },
   lState = {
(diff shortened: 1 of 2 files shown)


test/viewDirectiveSpec.js
```diff
@@ -324,7 +324,7 @@ describe('uiView', function () {
   });
 
   it('should instantiate a controller with controllerAs', inject(function($state, $q) {
-    elem.append($compile('<div><ui-view>{{vm.someProperty}}</ui-view></div>')(scope));
+    elem.append($compile('<div><ui-view></ui-view></div>')(scope));
```
jpekkala: This test broke but not because of what it tests. The template property was not originally in the test and after this change the template property overwrites the initial view.

christopherthielen: I'm not sure what this was originally trying to test >_>

PR #3165, opened 2016-11-21
feat(view): Route a view to a directive using `componentProvider`

sarahpads:
State definitions can dynamically load a component's template by using componentProvider.


christopherthielen:
very, *very* well done, thank you!!

Issue #3123, opened 2016-11-02
$templateFactory (0.3.1) is not using $templateRequest of AngularJS for templateUrl


anantanandgupta:
With the use of $templateRequest it is possible to have additional configuration while making the requests for the template. But since the ui-router's $templateFactory is not using that, it is not possible to have additional configuration for the ui-view's templateUrl.

In my project (ASP.Net MVC), all the views are coming from acition views and we need to pass the headers to all the requests for view. But it seems it is not possible if I am using templateUrl for the ui-view.

I have overcome the issue by decorating the factory and overwriting the implementation.

christopherthielen:
If you can get all the tests passing, Ill merge a PR.  The code should be compatible with angular 1.2 and fall back to `$http` if `$templateRequest` is not avilable, or the user chooses to disable it (via `$templateRequestProvider`)

See this previous PR #1900

anantanandgupta:
Hi Chris,

I will take some on that and will post a pull request. But I am not sure that I will be able to make it compatible to Angular 1.2.x.

One question though what do you mean by disabling $templateRequestProvider

christopherthielen:
Wonderful news!

ui-router supports ng 1.2.  The code should be compatible with 1.2.  It should fall back to `$http` if `$templateRequest` is not available (but use `$templateRequest` if it is available).  To achieve this, we can use `$injector.has` function to check if `$templateRequest` exists.  https://code.angularjs.org/1.2.0/docs/api/AUTO.$injector#methods_has

> One question though what do you mean by disabling $templateRequestProvider

I meant to type `$templateFactoryProvider`.  The user should also be able to opt-in to the old behavior (fetching templates using `$http`) if they choose.  We should create a `templateFactoryProvider` and add a config option.

anantanandgupta:
Hi @christopherthielen,

I have spent some time on this. The only restriction I face that there is a requirement of sending the requests with header `"Accept : text/html"`. I am not sure why that restriction is set. with `$templateRequest` there is no option to pass the configuration option via service call. It can only be configured in the config block as the global setting. So if we somehow move the responsibility of setting that header to the config section, as that if user requires it, he can set that, rather enforcing it. We can do this modification and make it support all the AngularJS versions (with of course fallback to `$http`).

Please suggest!

christopherthielen:
@anantanandgupta 

> So if we somehow move the responsibility of setting that header to the config section, as that if user requires it, he can set that, rather enforcing it. 

[...]

PR #3182, opened 2016-12-02
Fix for #3123 (thanks to @anantanandgupta): use $templateRequest in templateFactory

rjamet:
Based on PR #3155 from @anantanandgupta, fixing #3123 with more tests on top of it.

I've also updated the 1.3.x Angular used for tests to the latest one, 1.3.0 had a few $sce-related bugs. I got the Angular files from Google's CDN.

Finally, I'm not entirely sure this is the right way to send these commits, as I'm not too used to github: I can't propose commits to his PR, right? I'd be happy to send this another way if this is not the appropriate workflow :/


rjamet:
Oops, missed the gruntfile. Fixed.

anantanandgupta:
@rjamet Thanks for providing those tests. I guess that is no issue in that way of pull request as your commits are completely bases on my commits, if admins want to complete and pull my PR, they should be able to do it along with your commit.

christopherthielen:
Looks good.  Since this is a breaking change it will require a 0.4.0 release.
Before releasing that, I need a way to opt into the new $templateFactory behavior.
To do that, the service should be converted to a provider, which has `$get()` and has a `useTemplateRequest()` fn. 

Then users who do not want to keep using $http can call:

.config($templateFactory => $templateFactory.useTemplateRequest(false));`

Last but not least, this should be merged into master.

PR #3301, opened 2017-01-27
Negative parameters allowed in $state

### LINKED ISSUE #2422: State with negative int parameter fails to activate
When a state with an int parameter /url/{id:int} is activated where the value of the parameter is negative (-1) the state does not activate and the default route is activated. 

This works in version 0.2.13 but fails in the latest version of 0.2.15.

KatonRyu:
fix($state): Negative int parameters in $state allowed

Using a negative parameter in $state.go caused $rootScope:infdig errors; this is now fixed.

Closes #2422


christopherthielen:
Looks good, thanks!

PR #3330, opened 2017-02-09
fix(uiSref): replace angular.copy with extend

### LINKED ISSUE #3189: State transition breaks ES6 Map when passed through optional $stateParams in a ui-sref
When using an ES6 Map as a optional state param in order to pass data from one state to another, the Map object's methods stop working once passed through to another state.

http://plnkr.co/edit/TVANV1Cnb8B7h95mNjmW?p=preview

in the plunkr, transitioning between states without passing any stateParams works, and just instantiates a new Map.  If you click one of the links that passes the Map as a stateParam, the plunkr breaks with the error:

`Method Map.prototype.<methodName> called on incompatible receiver #<Map>`

mrohr:
using angular.copy broks states with optional params containing ES6 objects like Map and Set.  The rest of ui-router seems to use extend vs angular.copy, and a deep copy shouldn't be needed here.

this fixes #3189.

### DIFF
--- a/src/directives/stateDirectives.ts
+++ b/src/directives/stateDirectives.ts
@@ -249,10 +249,10 @@ uiSref = ['$uiRouter', '$timeout',
 
         if (ref.paramExpr) {
           scope.$watch(ref.paramExpr, function (val) {
-            rawDef.uiStateParams = angular.copy(val);
+            rawDef.uiStateParams = extend({}, val);
             update();
           }, true);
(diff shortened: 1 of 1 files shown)


christopherthielen:
This seems good.  The copy was introduced in b5f7b59692ce4933e2d63eb5df3f50a4ba68ccc0  to support references to objects in the expression.

I'm actually not sure that the extend (nor copy) is necessary either, but it shouldn't hurt.

christopherthielen:
Thanks for your contribution!

PR #3343, opened 2017-02-18
feat(uiSref): Bind ui-sref to DOM events

### LINKED ISSUE #3169: Bind ui-sref to DOM events other than click
There are occasions when `ui-sref` will not work as intended as when bound to the click event.  A common example is when a navigation menu that will hide when the blur event is triggered.  Blur is triggered before click so a `ui-sref` link will do nothing.  The [work around](http://stackoverflow.com/questions/27514036/using-ng-blur-and-ui-sref-doesnt-work-as-expected) is to execute `$state.go` on mousedown.  Other times, I want to highlight on a single click and go to a state on doubleclick.   

Could there be an option to override the default binding where mousedown and dblClick are possible   triggers for `ui-sref`?

gcca:
Using `ui-sref-opts="{ event: ['click', 'dblclick'] }"` we can override the default binding to `ui-sref` directive.

Fixes #3169

### DIFF
--- a/src/directives/stateDirectives.ts
+++ b/src/directives/stateDirectives.ts
@@ -12,7 +12,7 @@ import { ng as angular } from "../angular";
 import { IAugmentedJQuery, ITimeoutService, IScope, IInterpolateService } from "angular";
 
 import {
-    Obj, extend, forEach, tail, isString, isObject, parse, noop, unnestR, identity, uniqR, inArray, removeFrom,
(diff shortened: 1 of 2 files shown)


christopherthielen:
Looks good!  Could you please:

- also update the `ui-state` directive https://github.com/angular-ui/ui-router/blob/master/src/directives/stateDirectives.ts#L391-L396
- add some docs https://github.com/angular-ui/ui-router/blob/master/src/directives/stateDirectives.ts#L158-L161

gcca:
Ready.

PR #3390, opened 2017-03-26
use a true anonymous function inside inherit

cozmy:
This way, objects which inherit from a parent won't appear as type `extend` but of the generic type `Object`. I've only noticed this behaviour in Chrome's console.

Before:

image

After:

image

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -13,7 +13,7 @@ var isDefined = angular.isDefined,
     toJson = angular.toJson;
 
 function inherit(parent, extra) {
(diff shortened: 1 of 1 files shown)


christopherthielen:
Interesting.  Thanks!  

However, I have some concerns.  First, this broke the build by breaking jslint.  Second, I searched for information on this, and it seems that new Function() is currently an order of magnitude slower in Chrome.

http://jsperf.com/fastest-way-to-create-empty-function-in-javascript
https://github.com/jshint/jshint/issues/1619

PR #3397, opened 2017-03-30
bug(templates): `getTemplate` is returning undefined

amcdnl:
## Bug
In some cases, `cfg.getTemplate` is not defined  and throwing an error ( https://www.screencast.com/t/YmMUhXdByw ). The template is defined via `templateURL` like so: https://www.screencast.com/t/S7JKAslMB . 

Upon further debugging, in the follow screen https://www.screencast.com/t/qGGv9UXcsrV you can see the `data.$cfg` is an object, preventing the `noop` from being defined, however, the `getTemplate` function is not defined.

This fix adds a safety check to use `noop` if this is not defined.


senthanal:
👍

PR #3410, opened 2017-04-24
bug(directives): `data.$cfg` is passed but `getTemplate` is undefined

amcdnl:
## Issue
In some cases, `data.$cfg` is passed but `getTemplate` is still undefined...if thats the case, just return out. Setting `data.$cfg.getTemplate` to `noop` does not work in this case. I suspect this is a edge case caused by a hybrid environment.

## Solution
if `data.$cfg.getTemplate` is undefined, just return out of the loop. 

## Environment
- `ui-router` - `1.0.0-rc.1`
- `@ui-router/angular` - `1.0.0-beta.5`
- `@uirouter/core` - `5.0.0`
- `ui-router-ng1-to-ng2` - `2.0.0`

### DIFF
--- a/src/directives/viewDirective.ts
+++ b/src/directives/viewDirective.ts
@@ -362,6 +362,11 @@ function $ViewDirectiveFill($compile: angular.ICompileService,
         }
 
         let cfg: Ng1ViewConfig = data.$cfg || <any> { viewDecl: {}, getTemplate: noop };
+          
+        // in some cases, `data.$cfg` is passed but `getTemplate` is still undefined
+        // if thats the case, just return out. Setting `noop` does not work in this case.
+        if(!cfg.getTemplate) return;
+          
         let resolveCtx: ResolveContext = cfg.path && new ResolveContext(cfg.path);
         $element.html(cfg.getTemplate($element, resolveCtx) || initial);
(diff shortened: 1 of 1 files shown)


hoeni:
Had the same warning here. The patch worked fine for me.

Oliv:
The patch worked, but I don't know why in my case :

`typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('angular'))`

`ng_from_import`, meaning `require('angular')`, is an empty object.

If I replace :
`var cfg = data.$cfg || { viewDecl: {}, getTemplate: ng_from_import.noop };`
with : 
`var cfg = data.$cfg || { viewDecl: {}, getTemplate: angular.noop };`
it is working fine.

dunecatnat:
Hello! The fix works fine for me, but wanted to find out if there are plans to merge it and include in a release any time?

Root-Core:
Running into the same problem.. this should really be fixed. I don't know why the `getTemplate()` function is undefined in the first place or if this should ever happen. 

`angularjs@1.5`
`@uirouter/angularjs@1.0.10`

christopherthielen:
Fix coming soon, thanks for your patience

PR #3479, opened 2017-06-22
Fix ngStrictDi support

abyx:
In some scenarios, the `$injector` would be asked to annotate without
providing it with the current `strictDi` value, making resolvables not
complain about missing annotations.

This was previously addressed in this PR: https://github.com/angular-ui/ui-router/pull/2602

Yet, it is not working, as can be seen in this plunker, that contains a resolvable without annotations and yet no error is thrown: http://plnkr.co/edit/dd6ptmrUtzPnKLmAJkCY?p=preview


christopherthielen:
Thanks a lot!

Issue #3300, opened 2017-01-27
$state.is() is not working as expected from v0.4.0


Pmyl:
$state.is() when provided with the correct state and the _params_ parameter it will wrongly return **true** as long as the last _param_ matches even if the others are different.

This test should work but it fails:
```
it('should return false when the current state is passed with non matching parameters', inject(function ($state, $q) {
  $state.transitionTo(D, {x: 'foo', y: 'qux'});
  $q.flush();

  expect($state.is(D, {x: 'bar', y: 'qux'})).toBe(false);
}));
```

christopherthielen:
Thanks for the broken test!
I'll take a look and release 0.4.3 when it's fixed.

seavor:
Any update on when this will get patched?

pearcemerritt:
My team also ran into this issue, so I created a PR: https://github.com/angular-ui/ui-router/pull/3530

pearcemerritt:
@christopherthielen

PR #3530, opened 2017-09-14
fix(state): $state.is will wrongly return true

pearcemerritt:
as long as the last param matches even if the others are different.

Fixes https://github.com/angular-ui/ui-router/issues/3300

@christopherthielen


christopherthielen:
Thanks!

PR #3547, opened 2017-10-10
UI router init requires $uiRouter that requires $locationProvider (ng)

empyro:


empyro:
My integrated tests fails due to this bug.

PR #3556, opened 2017-11-01
Fix bower version reference

erickwilder:
The latest release was `1.0.10` and it still points to `1.0.8`


erickwilder:
I'm not sure if this is the proper way to solve this because there's a `bower_release.js` under `scripts/` and just reading the code I couldn't figure out what's wrong there (especially because the version there is based on the `package.json`file)

Is that file deprecated/not used anymore?

All in all I'm just following what was done [before](https://github.com/angular-ui/ui-router/commit/ce40527a062b50c0e5702589d29c61d3df723abe)

erickwilder:
Was the bower support dropped entirely?

christopherthielen:
sorry I missed this.  bower support is continuing, but seems to be broken.  I'll take a look at what's broken in the `bower_release.js` script

PR #3619, opened 2018-01-26
fix(uiSrefActive): don't match fuzzy on lazy loaded future states

fpipita:
This PR addresses an issue I've got into by working with lazy loaded future states and the `ng-{style,class}` interface. 

Basically, the directive didn't update its private list of the known states when future states were resolved so the `uiSrefActive` instances that were linked before the lazy states were added to the `stateRegistry`, kept incorrectly doing a fuzzy match on all the future state's lazily registred children.


christopherthielen:
thanks!

PR #3721, opened 2018-07-14
Add 'module' entry to allow Webpack to tree shake it

felixmosh:
Bundlers such as Webpack are not support `jsnext:main` entry, but do support `module`.

I've just added that entry, I've checked on my build, and webpack was able to enable module concat optimisation and tree shaking.


felixmosh:
@christopherthielen can you review that PR as well?

felixmosh:
@christopherthielen, ping

PR #3746, opened 2018-10-29
Angular-ui-router: Removed PhantomJS Dependency

ossdev07:
PhantomJS is an unmaintained project, removed phantomjs support from package.json.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>


rhenwood-arm:
I would like to see PhantomJS removed too. @christopherthielen can you provide feedback on this PR?

PR #3821, opened 2020-08-17
fix: drop ng_from_import to avoid webpack warning

zackschuster:
due to #3721 webpack warns that export 'module' from ng_from_import cannot be found. this PR removes the import statement, which removes the warning.


zackschuster:
this is no longer an issue for me & i have successfully updated my code from `1.0.19` to `1.0.28` 😄

PR #3838, opened 2021-11-08
Bump patch version on core

oBusk:
The version on @uirouter/angularjs is very strict and means you can't update to the latest patch version of @uirouter/core


christopherthielen:
I'm not comfortable changing how the dependencies of this library work, but we can bump to exactly 6.0.8 in #3840

